PR #5, opened 2023-09-10
feat(resize): resize

### LINKED ISSUE #3: [feat] add resize
Move `ngx-resize` to `ngxtension`

nartc:
closes #3 

This PR adds:
- `resize` entry point for resize utility using `ResizeObserver`
- a `test-app` with `test-app-e2e` for testing utilities that make sense to test visually (i.e: resize)
- `create-injection-token` entry point for `createInjectionToken()` utility


libs/ngxtension/create-injection-token/src/create-injection-token.ts
```diff
@@ -0,0 +1,153 @@
...
+			provider = { provide: token, useValue: value };
+		} else {
+			provider = {
+				provide: token,
+				useFactory: factory,
+				deps: (deps ?? []) as FactoryProvider['deps'],
+			};
+		}
+
+		return extraProviders ? [extraProviders, provider] : provider;
```
eneajaho, line 95: Do we care about cases where we want to create EnvironmentProviders ? 

Maybe another param? 
Would it be better if we add an object for options and not leave them as separate param ?

eneajaho (approved):
LGTM

libs/ngxtension/create-injection-token/src/create-injection-token.ts
nartc: what's the difference between Provider and EnvironmentProvider in terms of consuming it?

PR #39, opened 2023-09-14
feat(create-effect): add createEffect

### LINKED ISSUE #27: RFC 1
Don't you mind if I'll add this function?
https://gist.github.com/e-oz/62fed6d626df5fab5e34402b5f6ec06e
It's a standalone port of `ComponentStore.effect()` method, with 1 modification: it re-subscribes on errors.
Or you could add it yourself (it would be faster, I think), because the original function is not mine anyway so I'm not looking for any credit for that small modification :)

nartc:
closes #27


eneajaho:
LGTM!

I also had this copied somewhere in my codebase 😂. Glad we will have it here now!

eneajaho (approved):

PR #52, opened 2023-09-15
feat(navigation-end): add navigationEnd util function

### LINKED ISSUE #43: Add support function for router NavigationEnd event
Let's add an util function for catching whenever router navigation NavigationEnd event occur

e.g

```
export const routerNavigationEnd = () => {
  const router = inject(Router);

  return router.events.pipe(
    filter((event): event is NavigationEnd => event instanceof NavigationEnd)
  );
};
```

I am eager to add it if you would be interested with that one.

va-stefanek:
closes #43

### DIFF
--- a/docs/src/content/docs/utilities/navigation-end.md
+++ b/docs/src/content/docs/utilities/navigation-end.md
@@ -1,0 +1,35 @@
+---
+title: injectNavigationEnd
+description: ngxtension/navigation-end
+---
+
+The `injectNavigationEnd` function is a utility for creating an `Observable` that emits when a navigation ends. It might perform tasks after a route navigation has been completed.
+
+```ts
+import { injectNavigationEnd } from 'ngxtension/navigation-end';
+```
+
+## Usage
+
+`injectNavigationEnd` accepts optionally `Injector`.
+
+```ts
+import { Component } from '@angular/core';
+import { injectNavigationEnd } from 'ngxtension/navigation-end';
+import { NavigationEnd } from '@angular/router';
+
+@Component({
+	standalone: true,
+	selector: 'app-example',
+	template: '<p>Example Component</p>',
+})
+export class ExampleComponent {
+	navigationEnd$ = injectNavigationEnd();
+	constructor() {
+		navigationEnd$.subscribe((event: NavigationEnd) => {
+			// This code will run when a navigation ends.
+			console.log('Navigation ended:', event);
+		});
+	}
+}
+```
--- a/libs/ngxtension/navigation-end/src/navigation-end.spec.ts
+++ b/libs/ngxtension/navigation-end/src/navigation-end.spec.ts
@@ -1,0 +1,52 @@
+import { Component } from '@angular/core';
+import {
+	ComponentFixture,
+	TestBed,
+	fakeAsync,
+	tick,
+} from '@angular/core/testing';
+import { NavigationEnd, Router } from '@angular/router';
+import { delay, of } from 'rxjs';
+import { injectNavigationEnd } from './navigation-end';
+
+describe(injectNavigationEnd.name, () => {
+	@Component({
+		standalone: true,
+		template: '',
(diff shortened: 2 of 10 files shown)


libs/ngxtension/navigation-end/src/navigation-end.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Creates an Observable that emits when a navigation ends.
+ * @returns An Observable of NavigationEnd events.
+ */
+export function navigationEnd(injector?: Injector): Observable<NavigationEnd> {
+	injector = assertInjector(navigationEnd, injector);
+	return runInInjectionContext(injector, () => {
+		const router = inject(Router);
+
+		return router.events.pipe(
```
nartc, line 16:
```suggestion
		return inject(Router).events.pipe(
```

docs/src/content/docs/utilities/navigation-end.md
```diff
@@ -0,0 +1,35 @@
+---
+title: navigationEnd
+description: ngxtension/navigation-end
+---
+
+The `navigationEnd` function is a utility for creating an Observable that emits when a navigation ends. It might be used to perform tasks after a route navigation has been completed.
```
nartc, line 6:
```suggestion
The `navigationEnd` function is a utility for creating an `Observable` that emits when a navigation ends. It might perform tasks after a route navigation has been completed.
```

libs/ngxtension/navigation-end/src/navigation-end.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+			useValue: {
+				events: of(new NavigationEnd(0, '', '')),
+			},
+		});
+		fixture = TestBed.createComponent(Foo);
+		fixture.autoDetectChanges();
+		component = fixture.componentInstance;
+	});
+
+	it('should handle async stuff', fakeAsync(() => {
```
nartc, line 41: nit: maybe change the test name?

libs/ngxtension/navigation-end/src/navigation-end.ts
```diff
@@ -0,0 +1,22 @@
+import { Injector, inject, runInInjectionContext } from '@angular/core';
+import { Event, NavigationEnd, Router } from '@angular/router';
+import { assertInjector } from 'ngxtension/assert-injector';
+import { Observable } from 'rxjs';
+import { filter } from 'rxjs/operators';
```
nartc, line 5:
```suggestion
import { filter, type Observable } from 'rxjs';
```

nartc (changes requested):
It looks like you ran `pnpm exec nx g @nx/angular:entry-point`. You should run `pnpm exec nx g local-plugin:entry-point` instead. But it's ok, you can run `pnpm exec nx g local-plugin:convert-entry-point-to-project` generator to convert `navigation-end` to a separate project.

nartc:
@all-contributors please add @va-stefanek for code

libs/ngxtension/navigation-end/src/navigation-end.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Creates an Observable that emits when a navigation ends.
+ * @returns An Observable of NavigationEnd events.
+ */
+export function navigationEnd(injector?: Injector): Observable<NavigationEnd> {
```
eneajaho, line 11: Should we call it injectNavigationEnd ? 

Because that's what we do, we inject the router and return the navigation end. 

This way it will be understood correctly what it does. 

Just like we have injectResize. 

What do you think @nartc @va-stefanek ?

va-stefanek: That makes sense, let's go with It.

nartc (approved):

nartc:
Thank you for your contribution

PR #53, opened 2023-09-15
feat(call-apply): Implemented call apply Pipes

### LINKED ISSUE #35: [feat] add call - apply Pipes
Don't you mind if I'll add this pipes?
https://gist.github.com/dmorosinotto/c74f2c0d2e4cef4790815ec3b0684cc7
They are two "old" Pure pipes that I wrote to safe call "pure function" inside a template/component that helps solving problem of _"not calling function in **template**" _

dmorosinotto:
Hi I've implemented `CallPipe` and `ApplyPipe` + added some tests and docs (hope to make it in the right places)

@nartc I edited even the CONTRIBUTING.md just to fix a little typo for the nx command to create the secondary entry because I haven't the `pnx` command that you wrote, or maybe it's some kind of script/alias??
And added an short section about test (because I need even the nx command to run them 😉)  
Maybe another section to add is about where to put and how to write docs?

Closes #35

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ We will review your PR as soon as possible, and your contribution will be greatl
 Most likely, you'll need to create new secondary entry point to put the new utility in. To create entry point, use the following command:
 
 ```shell
-pnpm nx g local-plugin:entry-point <name-of-your-utility> --library=ngxtension --skip-module
+pnpm exec nx g local-plugin:entry-point <name-of-your-utility> --library=ngxtension --skip-module
 ```
 
 #### Please write some Tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,15 @@ We will review your PR as soon as possible, and your contribution will be greatl
 Most likely, you'll need to create new secondary entry point to put the new utility in. To create entry point, use the following command:
 
 ```shell
-pnx nx g local-plugin:entry-point <name-of-your-utility> --library=ngxtension --skip-module
+pnpm nx g local-plugin:entry-point <name-of-your-utility> --library=ngxtension --skip-module
+```
+
+#### Please write some Tests
+
+Try to cover your new contrib with some tests and make it pass running:
+
+```shell
+pnpm exec nx run ngxtension/<name-of-your-utility>:test
 ```
 
 ## Development Setup
--- a/docs/src/content/docs/utilities/call-apply.md
+++ b/docs/src/content/docs/utilities/call-apply.md
@@ -3,7 +3,7 @@ title: call apply Pipes
 description: ngxtension/call-apply
 ---
 
-`callPipe` and `applyPipe` are simple standalone pipes that simplify the calling of a PURE functions passing params to it, they take advantage of the "memoization" offerd by pure pipes in Angular, and enforces that you use them only with PURE functions (aka if you use this inside the body function they throws error!)
+`callPipe` and `applyPipe` are simple standalone pipes that simplify the calling of PURE functions passing params to it; they take advantage of the "memoization" offered by pure pipes in Angular, and ensure that you use them only with PURE functions (aka if you use this inside the body function they throw errors!)
 
 ```ts
 import { CallPipe, ApplyPipe } from 'ngxtension/if-validation';
@@ -11,7 +11,7 @@ import { CallPipe, ApplyPipe } from 'ngxtension/if-validation';
 
 ## Usage
 
-Both `CallPipe` and `ApplyPipe` need a PURE function or method to invoke (aka you can't use `this` in the function body), the difference between the two is only in that invocation order and that `|call` is sutable only for funciton with 1-param, instead `|apply` works for function with any number of params.
+Both `CallPipe` and `ApplyPipe` need a PURE function or method to invoke (aka you can't use `this` in the function body), the difference between the two is only in that invocation order and that `|call` is suitable only for function with 1-param, instead `|apply` works for function with any number of params.
 
 ```ts
 import { Component, ChangeDetectionStrategy } from '@angular/core';
@@ -26,11 +26,7 @@ import { CallPipe, ApplyPipe } from 'ngxtension/call-apply';
 
 		<b>call UTC: {{ now | call : ISOFormat }}</b>
 		<i>with apply: {{ ISOFormat | apply : now }}</i>
-		<p>{{ join | apply : 'Hello' : ' world' : '!' }}</p>
-
-		<!-- PARAMS ARE NOT STRICT TYPED SORRY :-( ANY IDEA? -->
-		<div>{{ join | apply : 41 : 1 }}</div>
-		<!-- PRINT: 42 -->
+		<p>{{ join | apply : 'Hello' : 'world' : '!' }}</p>
 
 		<!-- IF YOU UNCOMMENT NEXT LINE IT WILL THROW ERROR 
 		<h1>THIS IS NOT PURE: {{ updateClock | apply }}</h1>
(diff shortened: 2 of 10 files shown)


CONTRIBUTING.md
```diff
@@ -55,7 +55,15 @@ We will review your PR as soon as possible, and your contribution will be greatl
 Most likely, you'll need to create new secondary entry point to put the new utility in. To create entry point, use the following command:
 
 ```shell
-pnx nx g local-plugin:entry-point <name-of-your-utility> --library=ngxtension --skip-module
+pnpm nx g local-plugin:entry-point <name-of-your-utility> --library=ngxtension --skip-module
```
nartc, line 58:
```suggestion
pnpm exec nx g local-plugin:entry-point <name-of-your-utility> --library=ngxtension --skip-module
```

docs/src/content/docs/utilities/call-apply.md
```diff
@@ -0,0 +1,52 @@
+---
+title: call apply Pipes
+description: ngxtension/call-apply
+---
+
+`callPipe` and `applyPipe` are simple standalone pipes that simplify the calling of a PURE functions passing params to it, they take advantage of the "memoization" offerd by pure pipes in Angular, and enforces that you use them only with PURE functions (aka if you use this inside the body function they throws error!)
```
nartc, line 6:
```suggestion
`callPipe` and `applyPipe` are simple standalone pipes that simplify the calling of PURE functions passing params to it; they take advantage of the "memoization" offered by pure pipes in Angular, and ensure that you use them only with PURE functions (aka if you use this inside the body function they throw errors!)
```

docs/src/content/docs/utilities/call-apply.md
```diff
@@ -0,0 +1,52 @@
...
+
+```ts
+import { CallPipe, ApplyPipe } from 'ngxtension/if-validation';
+```
+
+## Usage
+
+Both `CallPipe` and `ApplyPipe` need a PURE function or method to invoke (aka you can't use `this` in the function body), the difference between the two is only in that invocation order and that `|call` is sutable only for funciton with 1-param, instead `|apply` works for function with any number of params.
```
nartc, line 14:
```suggestion
Both `CallPipe` and `ApplyPipe` need a PURE function or method to invoke (aka you can't use `this` in the function body), the difference between the two is only in that invocation order and that `|call` is suitable only for function with 1-param, instead `|apply` works for function with any number of params.
```

libs/ngxtension/call-apply/src/call-apply.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+@Pipe({
+	name: 'apply',
+	pure: true,
+	standalone: true,
+})
+export class ApplyPipe implements PipeTransform {
+	transform(fn: Function, ...args: any[]): any {
```
nartc, line 39:
```suggestion
	transform<TFunction extends (...args: any[]) => any>(fn: TFunction, ...args: Parameters<TFunction>): ReturnType<TFunction> {
```

nartc (changes requested):
You're right that `pnx` is my alias 😅

Thank you for your contribution. I have some suggestions

nartc:
@all-contributors please add @dmorosinotto for code

CONTRIBUTING.md
dmorosinotto: It works without the `exec` ?

docs/src/content/docs/utilities/call-apply.md
dmorosinotto: Sorry for my bad english...

libs/ngxtension/call-apply/src/call-apply.ts
dmorosinotto: Totally agree that this is a more strict type for the function and maybe enable TS inference, but II think that even adding this Angular will not "strict type-check" the pipes on template... I'll try and let's you know.

CONTRIBUTING.md
nartc: Yeah but I'd rather be consistent with the commands.

docs/src/content/docs/utilities/call-apply.md
nartc: No worries!

docs/src/content/docs/utilities/call-apply.md
dmorosinotto: and my fat fingers (typo)

libs/ngxtension/call-apply/src/call-apply.ts
dmorosinotto: And instead this will solve my NON strict-type check, great catch @nartc  
I tried in [stackblitz](https://stackblitz.com/edit/stackblitz-starters-grvarf?file=src%2Fcall-apply.ts,src%2Fmain.ts) your typed FN version emit an error (at build-time) if apply is called with wrong types... 
Don't see any "~swigel~" on template (maybe is the LanguageService implementation of stackblitz) or simply not supported case for Angular LSP, but at least we have more safe strict type-check!  

Do you want that I change even the Call sign allowing only function with 0-1 (param?: any) or even better `transform<T = any, R = any>(value: T, args?: (param?:T)=>R): R {`

dmorosinotto: And then I revisit the tests and docs to remove the "NON strict" sample ;-)

libs/ngxtension/call-apply/src/call-apply.ts
```diff
@@ -0,0 +1,44 @@
...
+			}
+	  );
+
+@Pipe({
+	name: 'call',
+	pure: true,
+	standalone: true,
+})
+export class CallPipe implements PipeTransform {
+	transform(value: any, args?: Function): any {
```
nartc:
```suggestion
	transform<TFunction extends (...args: any[]) => any>(value: Parameters<TFunction>[0], fn: TFunction): ReturnType<TFunction> {
```

We can use the same approach for `call`

dmorosinotto: there is no way to pass more than 1 param to call, because you do `{{ par | call: fn }}` I've just updated my PR with a commit containing your suggestion, let's me know if you need any other update...

libs/ngxtension/call-apply/src/call-apply.ts
```diff
@@ -23,7 +23,7 @@ const NOTHIS = !('Proxy' in window)
 	standalone: true,
 })
 export class CallPipe implements PipeTransform {
-	transform(value: any, args?: Function): any {
+	transform<T = any, R = any>(value: T, args?: (param?: T) => R): R {
```
nartc, line 26:
```suggestion
	transform<T = any, R = any>(value: T, args?: (param: T) => R): R {
```

If you make `param?: T` then it doesn't work because TypeScript will complain that `T | undefined is not assignable to T`

dmorosinotto: Sorry my bad, I've just discovered and fixed this problem in my [stackbliz repro]() and forget to port back it to the code before commit 😜

dmorosinotto: updated commit and PR 😉

nartc (approved):

nartc:
Looks like there's some timezone issue :D; probably better to convert back to UTC for the test?

dmorosinotto:
> Looks like there's some timezone issue :D; probably better to convert back to UTC for the test?

NOOOOOOoooooooo my Date PUN totally failed for timezone issue :neckbeard:  what a FAIL 👎 
I'll change the expect in the test, and commit the fix, sorry for that, I was thinking that initializing Date with all the parts will be timezone safe as UTC but obviously correctly handling Timezone is **NP-complete** _"hard problem"_ even worst than invalidate cache! 🤓

dmorosinotto:
@nartc we forgot to update the main README.md to point to call-apply docs 
Can you fix it in a future commit? or do you prefer that I do it right now?

Issue #33, opened 2023-09-13
[feat] add trackById and trackByProp directives


dmorosinotto:
Don't you mind if I'll add this function?
https://gist.github.com/dmorosinotto/8deeb5b2db6c6dfcb9a927268ef85745
They are two standalone directives to help/easy handle trackBy in *ngFor , the original code is inspired by this [article](https://medium.com/ngconf/make-trackby-easy-to-use-a3dd5f1f733b)

nartc:
Go for it!

dmorosinotto:
@nartc I'll start to implement this, now that I have some "practice" with the other PR 
I only have 1 question: what kind of test can I create for something like trackBy??
I'm not a text expert, but can't think of something "useful" to test if the directive trackById and trackByOther really works, because they will set an input for the ngForTrackBy input and alter the behavior of the renderer, but how I can test it? any suggestion?

PR #59, opened 2023-09-16
feat(track-id-prop): add trackById and trackByProp directives #33

dmorosinotto:
I've implemented the `trackById` and `trackByProp` directives as discussed in #33 
+ added some tests (but don't found a way to verify the real implementation detail of trackBy)
+ added docs to astro + updated the main README to add the missing link to utilities: call-apply and this new ones

### DIFF
--- a/libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
+++ b/libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
@@ -1,18 +1,12 @@
 //INSPIRED BY https://medium.com/ngconf/make-trackby-easy-to-use-a3dd5f1f733b
 import { NgForOf } from '@angular/common';
-import {
-	Directive,
-	Input,
-	Provider,
-	inject,
-	type NgIterable,
-} from '@angular/core';
+import { Directive, Input, inject, type NgIterable } from '@angular/core';
 
 @Directive({
 	selector: '[ngForTrackById]',
 	standalone: true,
 })
-export class NgForTrackByIdDirective<T extends { id: string | number }> {
+export class NgForTrackById<T extends { id: string | number }> {
 	@Input() ngForOf!: NgIterable<T>;
 	private ngFor = inject(NgForOf<T>, { self: true });
 
@@ -25,18 +19,15 @@ export class NgForTrackByIdDirective<T extends { id: string | number }> {
 	selector: '[ngForTrackByProp]',
 	standalone: true,
 })
-export class NgForTrackByPropDirective<T> {
+export class NgForTrackByProp<T> {
 	@Input() ngForOf!: NgIterable<T>;
 	private ngFor = inject(NgForOf<T>, { self: true });
 
-	@Input()
+	@Input({ required: true })
 	set ngForTrackByProp(trackByProp: keyof T) {
-		if (!trackByProp) return; //throw new Error("You must specify trackByProp:'VALID_PROP_NAME'");
+		if (!trackByProp) return;
 		this.ngFor.ngForTrackBy = (index: number, item: T) => item[trackByProp];
 	}
 }
 
-export const TrackByDirectives: Provider[] = [
-	NgForTrackByIdDirective,
-	NgForTrackByPropDirective,
-];
+export const TRACK_BY_DIRECTIVES = [NgForTrackById, NgForTrackByProp] as const;
--- a/libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
+++ b/libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
@@ -6,7 +6,7 @@ import { Directive, Input, inject, type NgIterable } from '@angular/core';
 	selector: '[ngForTrackById]',
 	standalone: true,
 })
-export class NgForTrackById<T extends { id: string | number }> {
+export class TrackById<T extends { id: string | number }> {
 	@Input() ngForOf!: NgIterable<T>;
 	private ngFor = inject(NgForOf<T>, { self: true });
 
@@ -19,7 +19,7 @@ export class NgForTrackById<T extends { id: string | number }> {
 	selector: '[ngForTrackByProp]',
 	standalone: true,
 })
-export class NgForTrackByProp<T> {
+export class TrackByProp<T> {
 	@Input() ngForOf!: NgIterable<T>;
 	private ngFor = inject(NgForOf<T>, { self: true });
 
@@ -30,4 +30,4 @@ export class NgForTrackByProp<T> {
 	}
 }
 
-export const TRACK_BY_DIRECTIVES = [NgForTrackById, NgForTrackByProp] as const;
+export const TRACK_BY_DIRECTIVES = [TrackById, TrackByProp] as const;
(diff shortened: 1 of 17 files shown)


nartc:
So for testing, we can leverage a symptom of missing `trackBy`: all children DOM nodes are re-drawn (re-rendered). IIRC, we can use `MutationObserver` API to test this behavior. Use the `test-app` + Cypress to test this on a real browser would be my suggestion.

dmorosinotto:
I read all the tests of trackBy in Angular and honestly I don’t get what they assert no warning message is tracked and the output text in the feature will be the same even if trackBy doesn’t exists or will not work - or I’m blind and miss understanding their test

For e2e test sorry @nartc  I’m not able to write and instrument it so if you know how to get there please help me to write - or simply accept that this is the correct implementation it’s a one liner directive that simple set the trackBy function 
I don’t mind testing internal behavior that even Angular tests don’t cover/verify (if my first sentence is correct)

nartc:
@dmorosinotto I added a new page for `track-by` and a simple e2e test. 

The page renders 3 `ul` off of the same `people` list:
- without track by
- with `trackById`
- with `trackByProp`

The page tracks the mutations of each `ul` using `mutationsLength` object. The page renders all 3 properties of `mutationsLength`.

The test asserts that the number of mutations match the behavior of without `trackBy` vs with `trackBy` (for both `trackById` and `trackByProp`)

dmorosinotto:
**WOW** @nartc I'll carefully read it and definitely has to learn something today🤓! 
Thanks for taking care of it, and invest so many times to show me this 🙏

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
```diff
@@ -0,0 +1,42 @@
...
+	Provider,
+	inject,
+	type NgIterable,
+} from '@angular/core';
+
+@Directive({
+	selector: '[ngForTrackById]',
+	standalone: true,
+})
+export class NgForTrackByIdDirective<T extends { id: string | number }> {
```
nartc, line 15: nit: what do you think of dropping `Directive` off of the name? Like `NgFor` and `NgIf`, they don't have the `Directive` appended.

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
```diff
@@ -0,0 +1,42 @@
...
+	constructor() {
+		this.ngFor.ngForTrackBy = (index: number, item: T) => item.id;
+	}
+}
+
+@Directive({
+	selector: '[ngForTrackByProp]',
+	standalone: true,
+})
+export class NgForTrackByPropDirective<T> {
```
nartc, line 28: nit: what do you think of dropping Directive off of the name? Like NgFor and NgIf, they don't have the Directive appended.

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
```diff
@@ -0,0 +1,42 @@
...
+	@Input() ngForOf!: NgIterable<T>;
+	private ngFor = inject(NgForOf<T>, { self: true });
+
+	@Input()
```
nartc, line 32: suggestion: maybe we can use `{required: true}` here

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
```diff
@@ -0,0 +1,42 @@
...
+	set ngForTrackByProp(trackByProp: keyof T) {
+		if (!trackByProp) return; //throw new Error("You must specify trackByProp:'VALID_PROP_NAME'");
```
nartc, line 34: comment: maybe with `{required: true}`, we don't need to check here or throw error instead of early return

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
```diff
@@ -0,0 +1,42 @@
...
+		this.ngFor.ngForTrackBy = (index: number, item: T) => item[trackByProp];
+	}
+}
+
+export const TrackByDirectives: Provider[] = [
```
nartc, line 39: suggestion: For this, I'd use `TRACK_BY_DIRECTIVES` instead

nartc, line 39: question: is `Provider[]` a correct type for this? These aren't Providers are they?

nartc (commented):
Mostly LGTM but I left some comments

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
dmorosinotto: Totally open to this idea, I've just follow the Angular "best-practices & default" but the name becomes too long, and this is why I decided to export even the TrackByDirectives to "simplify/short" import 😉

What names do you prefer @nartc 
`NgForTrackById`  and `NgForTrackByProp`  
or even shorter:  `TrackById` and `TrackByProp` ?

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
dmorosinotto: Same as above ;-)

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
dmorosinotto: If I don't remember incorrectly required was added in NG16 I don't remember/checked the minimum version for peer-dep of ngxtensions?  If it's just 16+ for me no problem at all to use the required input and drop additional check in the setter

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
dmorosinotto: I have only one concern (before {required}) and was that the user mis-use the directive writing only `trackByProp` forgetting to specify the value, so the setter will be called but with '' o undefined I don't know, so I decided to NOT throw but simply do nothing, at the end ngFor works without trackBy with a default behavior of tracking index, so if the setter is called with '', null, undefined I simply "opt-out" to trackBy and fallback to the default ngFor behavior without erroring.

I don't ever test what happen if {required} is set for Input, but the user specify and explicit value of `trackByProp:''` so if you don't mind I prefer to keep my _"defense programming"_ and check the value before using.

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
dmorosinotto: As for the naming of Directives, I'm totally open to change the name of the const array to all uppercase, or there is any other Angular / 3rd-party libs that uses different casing?

dmorosinotto: Yeah you are right! Provider is not the right type, It works just becouse Provider include Type<T>
I checked the types in angular and it's `imports?: (Type<any> | ReadonlyArray<any>)[]`

Do you prefer that I explicit the type as :Type<any>[]  or simply remove the TS type declaration and simply `export const XXX = [D1, D2] as const;` //This will infer the ReadonlyArray or Tuple I think

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
nartc: shorter is better

libs/ngxtension/trackby-id-prop/src/trackby-id-prop.ts
nartc: Let’s try as const

dmorosinotto:
Ok all refactor done and fixed docs + test + e2e pass 😉

nartc (approved):
Awesome!

dmorosinotto:
🎉

PR #68, opened 2023-09-18
fix(navigation): use autogenerate fonction to create navigation sidebar

tomalaforge:
add autogenerate to generate navigation sidebar. 

This way you just have to create your file and you do not forget to add it inside the conf.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,18 +37,7 @@ nx generate ngxtension:init
 
 <!-- UTILITIES:START -->
 
-- [assert-injector](https://ngxtension.netlify.app/utilities/assert-injector)
-- [call-apply](https://ngxtension.netlify.app/utilities/call-apply)
-- [computed-from](https://ngxtension.netlify.app/utilities/computed-from)
-- [connect](https://ngxtension.netlify.app/utilities/connect)
-- [create-effect](https://ngxtension.netlify.app/utilities/create-effect)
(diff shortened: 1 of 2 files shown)


nartc:
Does it automatically sort alphabetically?

tomalaforge:
It's order alphabetically depending on the file name, not on the title name. So if the title has the same name as the file, it will work

tomer953:
> Does it automatically sort alphabetically?

Yes :)

**[Sorting](https://starlight.astro.build/reference/configuration/)**
> Autogenerated sidebar groups are sorted by filename alphabetically. For example, a page generated from astro.md would appear above the page for starlight.md.



what about the readme? maybe remove the list? otherwise some CI needed to sort the list there

nartc:
Yeah probably remove the list from readme and just reference the documentation url there

tomalaforge:
What do you mean ? I didn't understand what you want me to do

nartc:
@tomalaforge we can turn on autogenerate for the Astro documentation site. There's another list of utilities in the root `README.md`. @tomer953 and I were talking about removing the list from `README.md` altogether and only add a URL to the documentation site on the `README.md` instead.

tomalaforge:
change made, but I think we can find a better way to express it.

nartc (approved):
We can always iterate

tomer953:
Please note that the new PR will cause installation to be before introduction.

Issue #41, opened 2023-09-14
rxjs Helper


tomalaforge:
-`filterArray` and `mapArray` to filter and map on array value => avoid doing the `obs$.pipe(map(arr => arr.filter(...)))`

- `filterNullUndefined` to filter null and undefined values on observable steams

dmorosinotto:
Can I suggest a different naming for _filterNullUndefined_ --> `filterNil`

tomalaforge:
@eneajaho @nartc what do you think?

nartc:
PRs are welcome!

tomer953:
Someone close ? :)

PR #70, opened 2023-09-18
Docs: improve rxjs helper docs

tomer953:
I found some small typos and decided to change a bit more if it’s ok.

#41

Ps, you did not update the readme and the docs with the new utils.


tomalaforge:
thanks for correcting my english. 

There is also the mapArray that might need to be improved. 

Yes, I didn't change the docs because of my other PR. But I might need to add them depending on what @nartc and @eneajaho want to do

nartc (approved):

PR #74, opened 2023-09-20
added reduce-array operator

tomer953:
same as `mapArray` but for `.reduce()` function

### DIFF
--- a/libs/ngxtension/reduce-array/src/reduce-array.spec.ts
+++ b/libs/ngxtension/reduce-array/src/reduce-array.spec.ts
@@ -1,0 +1,26 @@
+import { of } from 'rxjs';
+import { reduceArray } from './reduce-array';
+
+describe(reduceArray.name, () => {
+	const dummyObs = of([1, 2, 3]);
+
+	it('given an observable of [1,2,3] and a reducing function of summing elements, then result is an observable of 6', (done) => {
+		const result = dummyObs.pipe(reduceArray((acc, n) => acc + n, 0));
+
+		result.subscribe((r) => {
+			expect(r).toEqual(6);
+			done();
+		});
+	});
+
+	it('given an observable of [1,2,3] and a reducing function of summing elements with index, then result is an observable of 9', (done) => {
+		const resultWithIndex = dummyObs.pipe(
+			reduceArray((acc, n, i) => acc + n + i, 0)
+		);
+
+		resultWithIndex.subscribe((rI) => {
+			expect(rI).toEqual(9);
+			done();
+		});
+	});
+});
--- a/libs/ngxtension/reduce-array/src/reduce-array.spec.ts
+++ b/libs/ngxtension/reduce-array/src/reduce-array.spec.ts
@@ -2,24 +2,61 @@ import { of } from 'rxjs';
 import { reduceArray } from './reduce-array';
 
 describe(reduceArray.name, () => {
-	const dummyObs = of([1, 2, 3]);
+	const input$ = of([1, 2, 3]);
+	const emptyArray$ = of([]);
 
-	it('given an observable of [1,2,3] and a reducing function of summing elements, then result is an observable of 6', (done) => {
-		const result = dummyObs.pipe(reduceArray((acc, n) => acc + n, 0));
+	it('sums elements, result is 6', (done) => {
+		const result = input$.pipe(reduceArray((acc, n) => acc + n, 0));
 
-		result.subscribe((r) => {
-			expect(r).toEqual(6);
+		result.subscribe((x) => {
+			expect(x).toEqual(6);
 			done();
 		});
 	});
 
-	it('given an observable of [1,2,3] and a reducing function of summing elements with index, then result is an observable of 9', (done) => {
-		const resultWithIndex = dummyObs.pipe(
-			reduceArray((acc, n, i) => acc + n + i, 0)
+	it('sums elements with index, result is 9', (done) => {
+		const result = input$.pipe(reduceArray((acc, n, i) => acc + n + i, 0));
+
+		result.subscribe((x) => {
+			expect(x).toEqual(9);
+			done();
+		});
+	});
+
+	it('no initial value, sums elements, result is 6', (done) => {
+		const result = input$.pipe(reduceArray((acc, n) => acc + n));
+
+		result.subscribe((x) => {
+			expect(x).toEqual(6);
+			done();
+		});
+	});
+
+	it('no initial value, sums elements with index, result is 9', (done) => {
+		const result = input$.pipe(reduceArray((acc, n, i) => acc + n + i));
+
+		result.subscribe((x) => {
+			expect(x).toEqual(9);
+			done();
+		});
+	});
+
+	it('empty array observable, with initial value 0, result is 0', (done) => {
+		const result = emptyArray$.pipe(reduceArray((acc, n) => acc + n, 0));
+
+		result.subscribe((x) => {
+			expect(x).toEqual(0);
+			done();
+		});
+	});
+
+	it('empty array observable, no initial value, result is undefined', (done) => {
+		const result = emptyArray$.pipe(
+			reduceArray((acc, n: number) => (acc !== undefined ? acc + n : n))
 		);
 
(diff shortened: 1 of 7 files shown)


libs/ngxtension/reduce-array/src/reduce-array.ts
```diff
@@ -0,0 +1,9 @@
+import { map } from 'rxjs';
+
+export const reduceArray = <T, R>(
```
nartc, line 3: suggestion: for `reduce`, it's better to make `R = T` so if the consumers don't pass in initial value, then the return type is the type of the item in the Array. I.e: `nums.reduce((a, c) => a + c)`

```suggestion
export const reduceArray = <T, R = T>(
```

tomer953: yeap, also the initialValue should be optional. 
ill add it tomorrow since i also want to include unit test for that case

nartc:
@tomalaforge anyone can review. if you'd like to approve PR, I can invite you to the repo. Just shoot me a DM on twitter

tomer953:
I've updated the operator to make `initialValue` optional and added more tests; everything seems fine now.

However, a couple of notes:

1. Simply changing `R = T` and making `initialValue?: R` isn't sufficient. When calling `reduce(reduceFn, initialValue)` with `initialValue` set to `undefined`, the result is `NaN`. Therefore, we need to handle two separate cases: `reduce(reduceFn)` and `reduce(reduceFn, initialValue)`.

2. My TypeScript skills aren't advanced enough to properly handle these cases using type overloads, so I haven't managed to avoid using `any`.

3. While writing tests, I noticed that calling `reduce` on an empty array without an `initialValue` in JavaScript throws an error:

```js
[].reduce((acc, n) => acc + n);
// Uncaught TypeError: Reduce of empty array with no initial value
```
In my opinion, our `reduceArray` function should not throw an error in such cases but should return `undefined` instead. 
```js
  if (!array.length) {
      return undefined;
  }
```

libs/ngxtension/reduce-array/src/reduce-array.ts
```diff
@@ -0,0 +1,24 @@
...
+		// call reduce function with initialValue
+		if (initialValue !== undefined) {
+			return array.reduce(reduceFn, initialValue);
+		}
+		// no initialValue
+		else {
+			// Javascript throws error if array is empty: [].reduce((acc,n) => acc +n)
+			// avoid errors and return undefined
+			if (!array.length) {
+				return undefined;
```
tomalaforge, line 18: I think returning an empty array is prefer in that case. what do you think?

tomer953:
> I set any on the overloaded function, but that really doesn't matter. that's only to combine both functions.

yes with `any` I did managed to do so as well, I had few versions working fine with `any` and had to choose one, but I guess yours is the best so far

I'll commit this change

tomalaforge:
Seems correct for me 👍 But I'll wait for @eneajaho or @nartc to double check

tomalaforge (approved):

libs/ngxtension/reduce-array/src/reduce-array.ts
```diff
@@ -0,0 +1,33 @@
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+export function reduceArray<T>(
+	reduceFn: (acc: T, item: T, index: number) => T
+): (source: Observable<T[]>) => Observable<T>;
+
+export function reduceArray<T, R = T>(
+	reduceFn: (acc: R, item: T, index: number) => R,
+	initialValue?: R
```
nartc, line 10: suggestion: `initialValue` here doesn't have to be optional
```suggestion
	initialValue: R
```

libs/ngxtension/reduce-array/src/reduce-array.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+	it('empty array observable, no initial value, result is undefined', (done) => {
+		const result = emptyArray$.pipe(
+			reduceArray((acc, n: number) => (acc !== undefined ? acc + n : n))
+		);
+
+		result.subscribe((r) => {
+			expect(r).toBeUndefined();
+			done();
+		});
+	});
```
nartc, line 62: nit: this test looks a bit weird to me. Here's what I'd do

```suggestion
	it('empty array observable, no initial value, result is undefined', (done) => {
		let count = 0;
		const result = emptyArray$.pipe(
			reduceArray((_, n) => {
				count += 1;
				return n;
			})
		);

		result.subscribe((r) => {
			expect(r).toBeUndefined();
			expect(count).toEqual(0);
			done();
		});
	});
```

libs/ngxtension/reduce-array/src/reduce-array.ts
```diff
@@ -0,0 +1,33 @@
...
+	reduceFn: (acc: any, item: T, index: number) => any,
+	initialValue?: any
+): (source: Observable<T[]>) => Observable<any> {
+	return map((array: T[]) => {
+		// call reduce function with initialValue
+		if (initialValue !== undefined) {
+			return array.reduce(reduceFn, initialValue);
+		}
+		// no initialValue
+		else {
```
nartc, line 23: nit: unnecessary else

libs/ngxtension/reduce-array/src/reduce-array.ts
```diff
@@ -0,0 +1,33 @@
...
+export function reduceArray<T, R = T>(
+	reduceFn: (acc: R, item: T, index: number) => R,
+	initialValue?: R
+): (source: Observable<T[]>) => Observable<R>;
+
+export function reduceArray<T>(
+	reduceFn: (acc: any, item: T, index: number) => any,
+	initialValue?: any
+): (source: Observable<T[]>) => Observable<any> {
+	return map((array: T[]) => {
```
nartc, line 17: nit/suggestion: the `T` for the implementation with overloads is redundant. Clean them all up with `any`. However for this case, that might be too many `any`. So here's my suggestion:

```ts
type AnyArray = Array<any>;
type ReduceParameters = Parameters<AnyArray['reduce']>;

export function reduceArray<T>(
	reduceFn: (acc: T, item: T, index: number) => T
): (source: Observable<T[]>) => Observable<T>;

export function reduceArray<T, R = T>(
	reduceFn: (acc: R, item: T, index: number) => R,
	initialValue: R
): (source: Observable<T[]>) => Observable<R>;

export function reduceArray(
	reduceFn: ReduceParameters[0],
	initialValue?: ReduceParameters[1]
) {
	return map((array: AnyArray) => {
		// call reduce function with initialValue
		if (initialValue !== undefined) {
			return array.reduce(reduceFn, initialValue);
		}
		// no initialValue

		// Javascript throws error if array is empty: [].reduce((acc,n) => acc +n)
		// avoid errors and return undefined
		if (!array.length) {
			return undefined;
		}
		// if array is not empty, call the reduceFn without initialValue
		return array.reduce(reduceFn);
	});
}
```

nartc (commented):
Last couple of nits.

libs/ngxtension/reduce-array/src/reduce-array.ts
tomalaforge: I think any is easier to understand, and not a problem inside overloaded function. But you way is more the TS way but feel harder to fully get the why 😅

tomer953: what about:

`type Reduce = Array<any>['reduce'];`

then:
```ts
export function reduceArray(
	reduceFn: Parameters<Reduce>[0],
	initialValue?: Parameters<Reduce>[1]
) {
	return map((array: Array<any>) => {
```

?

nartc: I'm ok with that

tomer953: Review now…

nartc (approved):

eneajaho:
@all-contributors please add @tomer953 for reduce-array operator

eneajaho:
@all-contributors please add @tomer953 for code

PR #75, opened 2023-09-20
Additional Intl API helpers

### LINKED ISSUE #72: [RFC] Additional Intl API suggestion
I'm not sure if it's the case, but **it might be a good idea to add some additional Intl APIs**, such as  [ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat), [RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) and others.

I built a collection of [Intl pipes](https://github.com/wanoo21/ngx-intl-helper/blob/main/helpers/README.md) a while ago, to support a few of them, and I could add some here if that is the case 🙂.

wanoo21:
As we've talked, I'm pushing the new `Intl` pipes that I think could be useful.

These are:
* [Intl.DisplayNames](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) 
* [Intl.ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat)
* [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules)
* [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)
* [Intl.supportedValuesOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf)

I've written some tests that cover all helpers.

Fixes #72

### DIFF
--- a/libs/ngxtension/intl/src/display-names.pipe.ts
+++ b/libs/ngxtension/intl/src/display-names.pipe.ts
@@ -1,0 +1,81 @@
+import {
+	inject,
+	InjectionToken,
+	LOCALE_ID,
+	Pipe,
+	PipeTransform,
+	Provider,
+} from '@angular/core';
+
+type DisplayNamesOptions = Omit<Intl.DisplayNamesOptions, 'type'>;
+
+/**
+ * @internal
+ */
+const defaultOptions: DisplayNamesOptions = {
+	style: 'short',
+	localeMatcher: 'lookup',
+	fallback: 'code',
+};
+
+/**
+ * @internal
+ */
+const DISPLAY_NAMES_INITIALS = new InjectionToken<DisplayNamesOptions>(
+	'DISPLAY_NAMES_INITIALS',
+	{
+		factory: () => defaultOptions,
+	}
+);
+
+/**
+ * Provides a way to inject the options for the DisplayNamesPipe.
+ *
+ * @param options The options to use for the DisplayNamesPipe.
+ * @returns The provider for the DisplayNamesPipe.
+ */
+export function provideDisplayNamesOptions(
+	options: DisplayNamesOptions
+): Provider {
+	return {
+		provide: DISPLAY_NAMES_INITIALS,
+		useValue: { ...defaultOptions, ...options },
+	};
+}
+
+/**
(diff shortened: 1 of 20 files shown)


libs/ngxtension/intl/src/display-names.pipe.ts
```diff
@@ -0,0 +1,81 @@
...
+		try {
+			return new Intl.DisplayNames(locale || this.locale, {
+				...this.defaultOptions,
+				type,
+			}).of(code);
+		} catch (e) {
+			return code;
+		}
+	}
+}
```
tomalaforge, line 81: question: why can we not change the style of DisplayName with pipe argument as well, to make it more specific to each pipe and not having to override the provider ?

tomalaforge:
documentation is missing inside docs/src/content/docs/utilities folder

Otherwise nice pipes

wanoo21:
@tomalaforge didn't know I needed to add docs as well, not a problem, I'll add it!

libs/ngxtension/intl/src/display-names.pipe.ts
```diff
@@ -0,0 +1,84 @@
...
+		locale?: string | string[]
+	): ReturnType<Intl.DisplayNames['of']> {
+		try {
+			return new Intl.DisplayNames(locale || this.locale, {
+				...this.defaultOptions,
+				type,
+				...(style ? { style } : {}),
+			}).of(code);
+		} catch (e) {
+			return code;
```
nartc, line 81: nit: let's log the error so folks know what's going on?

nartc (commented):
LGTM. I have 2 comments that apply to all pipes. Feel free to ignore them but I'd like to discuss if possible

wanoo21:
Maybe we should consider a more universal approach to managing errors.

nartc:
@wanoo21 Yes, I agree. Let me put up an issue for error handling internally and we can discuss there.

nartc (approved):

PR #78, opened 2023-09-23
fix(create-injection-token): add injector to create injection token

nartc:


migrations.json
```diff
@@ -0,0 +1,20 @@
...
+		{
+			"cli": "nx",
+			"version": "16.8.2-beta.0",
+			"description": "Remove invalid options (strict, noInterop) for ES6 type modules.",
+			"factory": "./src/migrations/update-16-8-2/update-swcrc",
+			"package": "@nx/js",
+			"name": "16-8-2-update-swcrc"
+		}
+	]
+}
```
tomalaforge, line 20: This file doesn't need to be commited, doest it ?

tomer953: Yep, After the migrations have been run, nx no longer needs this file

eneajaho (approved):
LGTM

migrations.json
eneajaho: Yeah it needs to be removed!

PR #80, opened 2023-09-23
feat: added lazy service function, docs and tests

### LINKED ISSUE #44: [feat] add lazyService fn
```ts
export function lazyService<T>(loader: () => Promise<Type<T>>): Observable<T> {
  const injector = inject(Injector);

  return defer(() => {
    return loader().then((service) => injector.get(service));
  });
}
```

https://medium.com/itnext/lazy-loading-services-in-angular-what-yes-we-can-cfbaf586d54e

eneajaho:
closes #44

### DIFF
--- a/docs/src/content/docs/utilities/lazy-service.md
+++ b/docs/src/content/docs/utilities/lazy-service.md
@@ -1,0 +1,76 @@
+---
+title: lazyService
+description: ngxtension/lazy-service
+---
+
+`lazyService` is a helper function that allows us to lazily load a service.
+
+Lazy loading services is useful when we want to shrink the bundle size by loading services only when they are needed.
+
+```ts
+import { lazyService } from 'ngxtension/lazy-service';
+```
+
+:::tip[Inside story of the function]
+Read more here: [Lazy loading services in Angular. What?! Yes, we can.](https://itnext.io/lazy-loading-services-in-angular-what-yes-we-can-cfbaf586d54e)
+:::
+
+## Usage
+
+`lazyService` accepts a function that returns a `Promise` of the service. The function will be called only when the service is needed.
+
+It can be a normal dynamic import or a default dynamic import from a module.
+
+```ts
+const DataServiceImport = () => import('./data-service').then((m) => m.MyService);
+// or
+const DataServiceImport = () => import('./data-service');
+```
+
+Then, we can use `lazyService` to lazily load the service.
+
+```ts data.service.ts
+@Injectable({ providedIn: 'root' })
+export class MyService {
+	data$ = of(1);
+}
+```
+
+```ts test.component.ts
+const DataServiceImport = () => import('./data-service').then((m) => m.MyService);
+
+@Component({
+	standalone: true,
+	imports: [AsyncPipe],
+	template: '<div>{{data$ | async}}</div>',
+})
+class TestComponent {
+	private dataService$ = lazyService(DataServiceImport);
+
+	data$ = this.dataService$.pipe(switchMap((s) => s.data$));
+}
+```
+
+We can also use `lazyService` not in an injection context, by passing an injector to it.
+
+```ts test.component.ts
+const DataServiceImport = () => import('./data-service');
+
+@Component({
+	standalone: true,
+	template: '<div>{{data}}</div>',
+})
+class TestComponent implements OnInit {
+	private injector = inject(Injector);
+
+	data = 0;
+
+	ngOnInit() {
+		lazyService(DataServiceImport, this.injector)
+			.pipe(switchMap((s) => s.data$))
+			.subscribe((value) => {
+				this.data = value;
+			});
+	}
+}
+```
--- a/libs/ngxtension/lazy-service/src/lazy-service.spec.ts
+++ b/libs/ngxtension/lazy-service/src/lazy-service.spec.ts
@@ -1,0 +1,174 @@
+import { AsyncPipe } from '@angular/common';
+import {
+	ChangeDetectorRef,
+	Component,
+	inject,
+	Injectable,
+	Injector,
+	OnInit,
+	Type,
+} from '@angular/core';
+import {
+	ComponentFixture,
+	fakeAsync,
+	TestBed,
+	tick,
+} from '@angular/core/testing';
+import { catchError, of, switchMap } from 'rxjs';
+import { lazyService } from './lazy-service';
+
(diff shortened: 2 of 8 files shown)


wanoo21:
I think you should be using `nx g local-plugin:entry-point --skipModule` to generate the library entry point ;)

Good function btw!

libs/ngxtension/project.json
tomer953, line 1: Is it necessary?
The **/* catch them already

tsconfig.base.json
tomer953, line 1: Not related to your PR, but this file is not synced with all the utilities
( `reduceArray` that I PR’d for example not appear here)
Not sure how important is it

docs/src/content/docs/utilities/lazy-service.md
LcsGa, line 1: If I'm not wrong, updating primitives within a lazy loaded service like this is not possible. Maybe it could appear somewhere in the docs.

libs/ngxtension/project.json
eneajaho: Thanks

eneajaho:
Putting this here for more context: https://twitter.com/Jean__Meche/status/1708199444636688422 
Maybe we merge the impl or go with Mathieu's one ☝️

libs/ngxtension/lazy-service/src/lazy-service.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @example
+ * ```ts
+ * const dataService$ = lazyService(() => import('./data-service').then((m) => m.MyService));
+ * or
+ * const dataService$ = lazyService(() => import('./data-service'));
+ * ```
+ */
+export function lazyService<T>(
+	loader: () => Promise<Type<T>> | Promise<{ default: Type<T> }>,
```
JeanMeche, line 20: We might want to have `ProviderToken<T>` instead of `Type<T>` since a ProviderToken could come with its factory. wdyt ?

libs/ngxtension/lazy-service/src/lazy-service.ts
```diff
@@ -0,0 +1,39 @@
...
+	injector?: Injector
+): Observable<T> {
+	injector = assertInjector(lazyService, injector);
+
+	return runInInjectionContext(injector, () => {
+		return defer(() => {
+			return loader()
```
JeanMeche, line 27: While this function works fine, it kind of bothers me that we can't mock the loaded instance. 

See http://riegler.fr/blog/2023-09-30-lazy-loading-mockable For an implementation that provides suck mocking abilities !

libs/ngxtension/lazy-service/src/lazy-service.ts
```diff
@@ -0,0 +1,39 @@
...
+ * const dataService$ = lazyService(() => import('./data-service'));
+ * ```
+ */
+export function lazyService<T>(
+	loader: () => Promise<Type<T>> | Promise<{ default: Type<T> }>,
+	injector?: Injector
+): Observable<T> {
+	injector = assertInjector(lazyService, injector);
+
+	return runInInjectionContext(injector, () => {
```
JeanMeche, line 25: In which case is this necessary ?

libs/ngxtension/lazy-service/src/lazy-service.ts
```diff
@@ -0,0 +1,39 @@
...
+		return defer(() => {
+			return loader()
+				.then((serviceOrDefault) => {
+					if ('default' in serviceOrDefault) {
+						return injector!.get(serviceOrDefault.default);
+					}
+					return injector!.get(serviceOrDefault);
```
JeanMeche, line 32: If the service uses `DestroyRef.onDestroy()` it will never be called. 

Even if `injector` is a `NodeInjector`, this works only with `providedIn: root`. So it's the root injector  that will provide the `DestroyRef` (and thus never call `OnDestroy`. 

The solution would be to create an `EnvironmentInjector` that provides the class we just lazy-loaded. 

Something like : 
```
    if (!(injector instanceof EnvironmentInjector)) {
      // this is the DestroyRef of the component
      const destroyRef = injector.get(DestroyRef);

      // This is the parent injector of the injector we're creating
      const environmentInjector = injector.get(EnvironmentInjector);

      // Creating an environment injector to destroy it afterwards
      const newInjector = createEnvironmentInjector([type as Provider], environmentInjector);

      // Destroy the injector to trigger DestroyRef.onDestroy on our service
      destroyRef.onDestroy(() => {
        newInjector.destroy();
      });

      // We want to create the new instance of our service with our new injector
      injector = newInjector;
    }
```

libs/ngxtension/lazy-service/src/lazy-service.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+			fixture.detectChanges();
+			expect(fixture.nativeElement.textContent).toBe('');
+			tick(1);
+			fixture.detectChanges();
+			expect(fixture.nativeElement.textContent).toBe('1');
+		}));
+	});
+
+	describe('lazy loads a service not in injection context', () => {
+		@Component({ standalone: true, template: '<div>{{data}}</div>' })
```
JeanMeche, line 106: This tests succeeds even without the `runInInjectionContext`

libs/ngxtension/lazy-service/src/lazy-service.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @returns An observable of the service.
+ *
+ * @example
+ * ```ts
+ * const dataService$ = lazyService(() => import('./data-service').then((m) => m.MyService));
+ * or
+ * const dataService$ = lazyService(() => import('./data-service'));
+ * ```
+ */
+export function lazyService<T>(
```
JeanMeche, line 19: This function only currently only works with providedIn:'root' services.

libs/ngxtension/lazy-service/src/lazy-service.ts
eneajaho: When we want to use it in ngOnInit

JeanMeche: But we're only calling `injector.get` ? Does that still requires to be in an injector context ? (if the injected class uses inject for example ?)

eneajaho: You're right!

PR #81, opened 2023-09-26
feat: add a signal base hostBinding function

LcsGa:
As the title of the pull request suggests I added a `hostBinding` function that can be used to bind properties like `@HostBinding` would, but to signals (either writable or readonly).

As you would do to properties bound with the decorator, you are still able to set/update/mutate the signal if it is writable.

> This is my very first OSS contribution, I hope I didn't mess up :sweat_smile:

### DIFF
--- a/docs/src/content/docs/utilities/host-binding.md
+++ b/docs/src/content/docs/utilities/host-binding.md
@@ -1,0 +1,46 @@
+---
+title: hostBinding
+description: ngxtension/host-binding
+---
+
+`hostBinding` is a function that returns either a _writable_ or _readonly_ signal and binds the value held in that signal to the host property passed as the first argument like `@HostBinding` would do.
+
+```ts
+import { hostBinding } from 'ngxtension/host-binding';
+```
+
+## Usage
+
+With `@HostBinding` you can bind a color from a class property:
+
+```ts
+@Component({
+  standalone: true;
+  selector: 'my-component'
+  template: '...'
+})
+export class MyComponent {
+  @HostBinding('style.color') color = 'red';
+
+  updateColor(color: 'red' | 'blue') {
+    this.color = color;
+  }
+}
+```
+
+With `hostBinding` you can now bind anything like `@HostBinding` on writable or readonly signals:
+
+```ts
+@Component({
+  standalone: true;
+  selector: 'my-component'
+  template: '...'
+})
+export class MyComponent {
+  color = hostBinding('style.color', signal('red'));
+
+  updateColor(color: 'red' | 'blue') {
+    this.color.set(color);
+  }
+}
+```
--- a/libs/ngxtension/host-binding/src/host-binding.ts
+++ b/libs/ngxtension/host-binding/src/host-binding.ts
@@ -1,0 +1,76 @@
+/* eslint-disable @typescript-eslint/ban-types */
+import {
+	effect,
+	ElementRef,
+	HostBinding,
+	inject,
+	Injector,
+	Renderer2,
+	RendererStyleFlags2,
+	runInInjectionContext,
+	Signal,
+	WritableSignal,
+} from '@angular/core';
+import { assertInjector } from 'ngxtension/assert-injector';
+
+/**
+ * `hostBinding` takes a `hostPropertyName` to attach a data property, a class, a style or an attribute (as `@HostBinding` would) to the host.
+ * The udpate is applied based on the update of the provided signal (writable or not).
+ *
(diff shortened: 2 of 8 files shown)


wanoo21:
It's so cool! I was just thinking about a function like this a few days ago. This is an awesome PR.

LcsGa:
> It's so cool! I was just thinking about a function like this a few days ago. This is an awesome PR.

Glad you like it 😄!

tomalaforge:
nice one 👍
well documented and well tested.

LcsGa:
> nice one 👍
> well documented and well tested. 

Thanks! 😁

docs/src/content/docs/utilities/host-binding.md
```diff
@@ -0,0 +1,46 @@
...
+  standalone: true;
+  selector: 'my-component'
+  template: '...'
+})
+export class MyComponent {
+  color = hostBinding('style.color', signal('red'));
+
+  updateColor(color: 'red' | 'blue') {
+    this.color.set(color);
+  }
```
nartc, line 44: question: can we add a usage in combination with Input? There will be 2 cases:
- Observable Input like people do today with setter Inputs

```ts
export class MyComponent {
    #color = new BehaviorSubject('red');
    @Input() set color(color: 'red' | 'blue') {
        this.#color.next(color);
    }    
    // how to use hostBinding() in this case
}
```

- Future _theoretical_ signal inputs

```ts
export class MyComponent {
    color = input('red');
    constructor() {
        // maybe this is how we'd use this.
        hostBinding('style.color', this.color);
    }
}
```

Without Signal Input, can we also add a usage with Input?

```ts
export class MyComponent {
    #color = hostBinding('style.color', signal('red'));
    
    @Input() set color(color: 'red' | 'blue') {
        this.#color.set(color);
    }
}
```

libs/ngxtension/host-binding/src/host-binding.ts
```diff
@@ -0,0 +1,73 @@
...
+						renderer.addClass(element, property);
+					} else {
+						renderer.removeClass(element, property);
+					}
+				},
+			})[binding] ?? (() => renderer.setProperty(element, binding, value))
+		)();
+	});
+
+	return signal;
```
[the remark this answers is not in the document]
LcsGa: Good catch about the injection context issue, I missed it 😅

About an eventual cleanup, I'm not 100% but I don't think there is anything to add! I'll investigate further though 😁

docs/src/content/docs/utilities/host-binding.md
LcsGa: About observables I thought about maybe creating a custom operator instead, but I'm not sure! What do you think?

I don't really know how signal based input will work but `hostBinding` could work the exact same way that of any `Signal`

About your last use case, it will work exactly as you wrote it 😁

libs/ngxtension/host-binding/src/host-binding.ts
```diff
@@ -0,0 +1,75 @@
...
+					attr: () => renderer.setAttribute(element, property, String(value)),
+					class: () => {
+						if (value) {
+							renderer.addClass(element, property);
+						} else {
+							renderer.removeClass(element, property);
+						}
+					},
+				})[binding] ?? (() => renderer.setProperty(element, binding, value))
+			)();
```
nartc, line 70: nit: I think this is trying to be too smart here. Does this have to be recreated every time the effect triggers? I would rather create the object outside of the `effect` or just good ol' if/elseif instead.

LcsGa: Well, you are right. I really like this syntax that has the exact same perfs as a switch/case (when not in a loop) but here it makes it less readable. Therefore I should stick with the old good switch/case (I wish that we will have the same one as java someday :stuck_out_tongue:)

About perfs it should not make such a big diff here, though. Even if it is within an effect (thus, a "loop"), variables with host binding linked to them are (usually ?) updated a few times. **But** we never know, so once again, I should refactor it to use a switch/case instead!

nartc (approved):
LGTM. Can you rebase?

nartc:
@all-contributors please add @LcsGa for code

LcsGa:
> LGTM. Can you rebase?

Yes I can! But do you mean rebasing in order to have a single commit or to resolve conflicts or both (I use merge way more than rebase, except for when I want a clean branch, so I prefer asking 😅)?

nartc:
@LcsGa both

LcsGa:
@nartc Done! I hope I didn't do it wrong :sweat_smile:

nartc:
@LcsGa Thanks for the contribution!

LcsGa:
> @LcsGa Thanks for the contribution!

With pleasure :smile:

PR #92, opened 2023-10-04
feat(gestures): port Drag Gesture from use-gesture

### LINKED ISSUE #91: [feat] add use-gesture/drag
https://use-gesture.netlify.app/docs/#vanilla-javascript

nartc:
Closes #91

### DIFF
--- a/libs/ngxtension/gestures/src/index.ts
+++ b/libs/ngxtension/gestures/src/index.ts
@@ -1,0 +1,2 @@
+export * from './drag';
+export * from './zoneless-gesture';
--- a/apps/test-app-e2e/src/e2e/drag.cy.ts
+++ b/apps/test-app-e2e/src/e2e/drag.cy.ts
@@ -1,0 +1,23 @@
+describe('drag', () => {
+	beforeEach(() => cy.visit('/drag'));
+
+	it('should update coord after drag', () => {
+		cy.get('.pink-draggable-box > span').as('coordSpan');
+		cy.get('.pink-draggable-box').as('box');
+		cy.get('@coordSpan').should('contain.text', '0,0');
+
+		cy.get('@box').trigger('pointerdown', {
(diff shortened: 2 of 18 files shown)


libs/ngxtension/if-validator/src/if-validator.ts
```diff
@@ -1,9 +1,9 @@
 import {
 	AbstractControl,
-	AsyncValidatorFn,
 	FormControl,
-	ValidatorFn,
 	Validators,
+	type AsyncValidatorFn,
+	type ValidatorFn,
```
tomalaforge, line 6: nitpick: this refacto should have been done in a separate PR

libs/ngxtension/gestures/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './drag';
+export * from './zoneless-gesture';
```
tomalaforge, line 2: personal opinion: I always prefer to explicitly export what I want to export

export {xxx} from '....'

tomalaforge:
you want to implement the other behavior in the same PR?

libs/ngxtension/if-validator/src/if-validator.ts
nartc: Yeah, makes sense. I'll turn on `verbatimModuleSyntax` in a different PR

libs/ngxtension/gestures/src/index.ts
nartc: I do it on a case-by-case basis. There's no value to do `export { xxx }` in these cases

nartc:
@tomalaforge I guess it doesn't have to be the same PR. I'll clean up the verbatim tsconfig then will update the PR with tests.

eneajaho:
LGTM! 

Can't wait to have all the helpers in the lib!

PR #98, opened 2023-10-05
fix(create-injection-token): allows multi token to work correctly

### LINKED ISSUE #95: [RFC] createInjectionToken with support of multi: true
I started to rewrite some of my libraries and added ngxtension.

I'm using [createInjectionToken](https://ngxtension.netlify.app/utilities/injectors/create-injection-token/) quite often and I've noticed I can't create a multi-token - `multi: true`.

I checked the source code and I don't see to be a way to mark it as multiple.

image

Maybe there's a workaround I can't find out?

nartc:
closes #95


tomalaforge:
That's nice, but I think a small chapter on the documentation page would be nice

tomalaforge (approved):

PR #100, opened 2023-10-06
fix(create-injection-token): allow provideFn to accept factory

### LINKED ISSUE #99: [RFC] createInjectionToken generic type and optional factory
I'd like to provide some other suggestions.

How about adding support for the **Generic type and marking the factory as optional**?

So basically this:
`export const TOKEN = new InjectionToken<MyType>();`

becomes:
`export const [injectFn, provideFn] = createInjectionToken<MyType>();`

There's a problem though, I can't make it `multi` without providing the factory 🤷🏻‍♂️

Also, considering the code above, would this work?
`provideFn(() => {
  const foo = inject(FOO);
  if (foo) return 'something'
  return 'something else'
})`

Instead of a simple value, to provide a factory!

This is how I manage it now:
image

nartc:
Closes #99

No docs yet because my brain is not firing on all cylinders this morning (Insomnia struck last night so I've been up since 2AM). I'll add docs later or @wanoo21 can put some docs 😋. 

cc @wanoo21 no docs so you might want to take a look at the new tests in this PR for usage.

For "optional factory" case, I had to add a new function `createNoopInjectionToken()` instead. Typing is tricky with the optional factory.

```ts
// "multi" token; unfortunately until I can acquire better typescript skill, we have to duplicate "true" here
const [injectFn, provideFn] = createNoopInjectionToken<number, true>('description', { multi: true });

injectFn(); // number[]
provideFn(1); // accepts number
provideFn(() => 1); // accepts a factory returning a number

// normal token
const [injectFn, provideFn] = createNoopInjectionToken<number>('description');
injectFn(); // number;
```

### DIFF
--- a/libs/ngxtension/create-injection-token/src/create-injection-token.spec.ts
+++ b/libs/ngxtension/create-injection-token/src/create-injection-token.spec.ts
@@ -1,6 +1,9 @@
 import { Injector, inject } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
-import { createInjectionToken } from './create-injection-token';
+import {
+	createInjectionToken,
+	createNoopInjectionToken,
+} from './create-injection-token';
 
 describe(createInjectionToken.name, () => {
 	describe('given root injection token', () => {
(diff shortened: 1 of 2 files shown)


wanoo21:
So fast, thanks @nartc, I'll check the source code and test examples this weekend, and I sure can help with docs!

You should take a break, hahaha!

eneajaho (approved):

nartc:
Waiting on @wanoo21 confirmation before merging.

wanoo21:
I think we're good to move. Let's merge it and I'll update the docs before release.

Should I create a new issue for documentation?

Issue #34, opened 2023-09-13
RFC - do you mind adding some helpers related to Observables and or TypeScript types?


dmorosinotto:
I think that this repo/project is mostly related to *Angular* helpers, but do you mind if I add something related to Observable operators: https://gist.github.com/dmorosinotto/f98370a39d9041358915b8a493cf3654
or TypeScript generic types helpers? 
https://gist.github.com/dmorosinotto/394ad458bd5352c3869981771baeaf1d 
I have even others but just asking what you think before opening proposal issues that maybe are not in scope/desired...

nartc:
I think the RxJS operators make sense to include but the TypeScript helpers aren't a bit of an outlier. Thoughts?

dmorosinotto:
Yes I suspect it, maybe I pick one sample types that is not really related to "Angular" I have other Generic types that I wrote and use in different project, but maybe it's out of scope for this library I agree

dmorosinotto:
For the Observables operator if you think that they be acceptable I will go to separate issue (feat add proposal) so you can evaluate it separately

nartc:
Separate issue would be great. And you can close this one with a reference to the other one. (or I can close it once you have a new issue up)

dmorosinotto:
@nartc sorry for the delay but was so busy in the latest week, I'll probably had some free times on this weekend to implements the RxJS operators that I mentioned, but before opening a PR I want to discuss with you and others @tomer953, @JustroX, @tomalaforge that had running issues about adding RxJs operators if we'll keep creating separate entry point one for each operator, or if we can take a different approach and organize all of them under a common `ngxtension/rxjs`  or `ngxtension/operators` and put all of them under this entrypoint as single export functions?  What do you all think?

tomer953:
Pretty sure @nartc has referred to this somewhere in the issues, the “flatten” approach is better duo to the secondary entry point feature, if you go one more level, another package will be needed to be a wrapper. (not sure at all)


> The library is entirely consisted of [Secondary Entry Point](https://angular.io/guide/angular-package-format#entrypoints-and-code-splitting). Even though we ship ngxtension as one package (so it is easy for consumers to install), Angular build pipeline should handle code-splitting and tree-shaking properly for all the entry points that ngxtension comes with.

dmorosinotto:
[...]

PR #101, opened 2023-10-07
feat (mapfilter) - Add mapFilter operator + docs

dmorosinotto:
I added the mapFilter operator proposed on #34 
after that I'll create a separate PR for debug operator.
The commit also added the mapfilter.md for docs, don't found the astro config for link it, maybe now is it all automated!?


docs/src/content/docs/utilities/Operators/mapfilter.md
```diff
@@ -0,0 +1,29 @@
+---
+title: mapFilter
+description: An RxJS operator that allow to apply a trasform function to each value of the observable in (same as map), but with the ability to skip (filter out) some values if the function explict return undefined or simply doesn't return anything for same code-path (implict return undefined).
```
ajitzero, line 3:
```suggestion
description: An RxJS operator that allows applying a transform function to each value of the observable in (same as map), but with the ability to skip (filter out) some values if the function explicit return undefined or simply doesn't return anything for same code-path (implicit return undefined).
```
Minor typos/grammar

docs/src/content/docs/utilities/Operators/mapfilter.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Import
+
+```typescript
+import { mapFilter } from 'ngxtension/mapfilter';
+```
+
+## Usage
+
+You can use it as a normal map operator, but with the ability to skip some values: returning undefined (explict or implict).
```
ajitzero, line 14:
```suggestion
You can use it as a normal map operator, but with the ability to skip some values: returning undefined (explicit or implicit).
```

libs/ngxtension/mapfilter/project.json
```diff
@@ -0,0 +1,33 @@
+{
+	"name": "ngxtension/mapfilter",
```
ajitzero, line 2: nit: Not sure about the naming scheme but should the file be `ngxtension/map-filter` instead (hyphen in middle?)

dmorosinotto: I don't know if there is an enforced/preferred naming convention for the files...

dmorosinotto:
@nartc about the failing CI I need to write the @tsignore inside the TEST to prove that the mapFunction correctly handle the case when someone don't produce/return any value for some code-path.
Locally the test pass, but not in CI can you please help me or suggest a valid workaround? Thanks

docs/src/content/docs/utilities/Operators/mapfilter.md
```diff
@@ -0,0 +1,29 @@
+---
+title: mapFilter
+sdescription: An RxJS operator that allows applying a transform function to each value of the observable in (same as map), but with the ability to skip (filter out) some values if the function explicit return undefined or simply doesn't return anything for same code-path (implicit return undefined).
```
ajitzero, line 3:
```suggestion
description: An RxJS operator that allows applying a transform function to each value of the observable in (same as map), but with the ability to skip (filter out) some values if the function explicitly returns undefined or simply doesn't return anything for same code-path (implicit return undefined).
```
One more

libs/ngxtension/mapfilter/src/mapfilter.ts
```diff
@@ -0,0 +1,11 @@
+import { type Observable } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
+
+export function mapFilter<T, R>(fnTrasformSkipUndefined: (value: T) => R) {
+	return function (source: Observable<T>): Observable<Exclude<R, undefined>> {
+		return source.pipe(
+			map(fnTrasformSkipUndefined),
+			filter((value) => value !== undefined)
```
tomalaforge, line 8: you can use `filterNil()`here, this way I don't think you need to cast you observable at the end.

libs/ngxtension/mapfilter/src/mapfilter.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+			})
+		);
+
+		let out = '>';
+		out$.subscribe((s) => {
+			//s is a string
+			out += s + '-';
+		});
+		out += '|';
+		expect(out).toEqual('>2-6-|');
```
tomalaforge, line 20: I don't get your test. Why are you not simply testing the returned value ? It makes your test very hard to understand. 
We could add observableSpy to eaiser test observable. What do you thing @nartc ?

libs/ngxtension/mapfilter/project.json
dmorosinotto: Ok @nartc I'll rename all files in next commit
PS: Do you have any suggestion about the @ts-ignore problem for CI?

libs/ngxtension/mapfilter/src/mapfilter.ts
dmorosinotto: @tomalaforge sorry I can't use your filterNil implementation because I want to filter ONLY undefined, null is a valid return value that must propagate through the filter, and so even the explicit cast at the end is needed.

libs/ngxtension/mapfilter/src/mapfilter.spec.ts
dmorosinotto: @tomalaforge I'm not an test expert expecially for RxJS, don't know how to use/rewrite it using observableSpy, in either case all my test passes! 
And my code is simply composing an out string that "mimic" the "marble diagram" of the out$, because I need to test the result checking the stream of events that occurs, and the single value that can pass through mapfilter (odd numbers -> 2x) and which one are filtered (even number)

dmorosinotto: Even if I use for semplicity a "sync" observable alle the value 1, 2, 3 are emitted as single number, notice I use in$= from([1,42,3]) that will produce an Observable with a "marble" representation like  this `>1-42-3-|` where 
`>` represents the start of the stream
`-` represents the time that sepatete one next from the subsequent
`|` is the complete event (end of the stream) 
`X` here not used, represents and error emission (that maybe end the stream)

Hope this will clarify even my test cases

tomalaforge: Yes I get the 'look like' marble test. But real marble test are hard to understand, so similar one are even harder. It's hard to get why you test it like that. 

My propose solution:
```ts
	it('given an observable >1-42-3| and a mapping function that double ONLY the odds value, then result is an observable of >"2"--"6"-| the intial even value (42) is not mapped and so filtered out', fakeAsync(() => {
		const in$ = of(1, 42, 3);
		const out$ = in$.pipe(
			mapFilter((n) => {
				if (n % 2) return String(n * 2);
				else return undefined; // explict return undefined to skipout (filter) some value from the out observable
			})
		);

		// let out = '>';
		out$.pipe(toArray()).subscribe((s) => {
			//s is a string
			expect(s).toEqual(['2', '6']);
		});
		tick();
		// out += '|';
		// expect(out).toEqual('>2-6-|');
	}));
```

or 
```ts
	it('given an observable >1-42-3| and a mapping function that double ONLY the odds value, then result is an observable of >"2"--"6"-| the intial even value (42) is not mapped and so filtered out', (done) => {
		const in$ = of(1, 42, 3);
		const out$ = in$.pipe(
			mapFilter((n) => {
				if (n % 2) return String(n * 2);
				else return undefined; // explict return undefined to skipout (filter) some value from the out observable
			})
		);

		// let out = '>';
		out$.pipe(toArray()).subscribe((s) => {
			//s is a string
			expect(s).toEqual(['2', '6']);
			done();
		});
		// out += '|';
		// expect(out).toEqual('>2-6-|');
	});
```

libs/ngxtension/mapfilter/src/mapfilter.ts
dmorosinotto: Because this is **MINE** desired implementation, and if you read the docs description: An RxJS operator that allows applying a transform function to each value of the observable in (same as map), but with the ability to skip (filter out) some values if the function explicitly returns undefined or simply doesn't return anything for same code-path (implicit return undefined).

I clarify that only undefined is filter out never spoken about null.
And the motivation is really simple even if "null+undefined" is sometimes called the "2 billion$ mistake" for me it's really clear that in JS undefined is a way of the JS Engine/runtime to express missing value, null is our (dev-world) to express missing value.

The intent of mapFilter is just to give to the dev-user a way to write an imperative trasform that choose which value to return (even null if they want, it's always a valid value in JS/TS) + skips some value emit to the down streams, in 2 simple ways: explicit return undefined, or either write a transform function that for some "code-path" doesn't return anything --> so the function for JS implicit return undefined (AND NOT null) and this is why I choose to filter out ONLY undefined.

If someone still want to filter out even null, can simply done `in$.pipe(mapFilter(...), filterNil())`

libs/ngxtension/mapfilter/src/mapfilter.spec.ts
dmorosinotto: Thanks for your suggested Test code, I've never known/used `toArray()` operator, quite un-usual one, but in this case can simplify the test expect. 
If you don't mind rewrite them I prefer to keep the "marble" representation...
PS: with **of**`([1,42,3])` the test will fail, you must use **from**`([1,42,3])`;

libs/ngxtension/mapfilter/src/mapfilter.ts
tomalaforge: Never wanted to upset you, that's why I added "question" in front. I was just asking a question. I'm out of this discussion. I leave @nartc or @eneajaho handle this.

dmorosinotto: @tomalaforge sorry for the tone, maybe too rude  
I didn't got your question intent initially... sorry 😞! 
Hope you can understand the motivation way I want to keep null, and filter only undefined.
Have a nice day/night

libs/ngxtension/mapfilter/src/mapfilter.spec.ts
nartc: I second the `observer-spy` approach. The tests work but they're hard to parse what's going on.

libs/ngxtension/mapfilter/src/mapfilter.ts
nartc: I agree with @tomalaforge. In my opinion, the value proposition of `mapFilter` comes from the narrower type that `filter((value) => value !== undefined` provides via `Exclude<R, undefined>`. 

`mapFilter` as-is is too limiting. If we say `mapFilter`, I'd expect to be able to pass in a `filterFn` as well as the `transformFn`. 

My proposal is one of the following:
1. rename to `mapFilterUndefined` or `mapSkipUndefined`
2. pivot to a different operator `filterUndefined` or `skipUndefined` to only provide the type-narrow feature
```ts
$.pipe(
    map(mapFn),
    filterUndefined() // same as filter(val => val !== undefined) but it is more strongly typed
)
```

libs/ngxtension/mapfilter/project.json
nartc: You can add `noImplicitReturns: false` to `tsconfig.spec.json`

libs/ngxtension/mapfilter/src/mapfilter.spec.ts
dmorosinotto: Is this [observer-spy](https://www.npmjs.com/package/@hirez_io/observer-spy) you are referring?
Do I can install it? And have to update the jest.config to handle auto-unsubscribe?!

Please @nartc can you suggest me how to solve the problem with CI failing because I used the @ts-ignore to write the test for empty code-path that doesn’t return value, or I simply remove the test before committing…

libs/ngxtension/mapfilter/src/mapfilter.ts
dmorosinotto: I don't think that people need a new operator to simply pipe(map, filter) if they need to pass in all the fnTrasform and fnFilter, so I'll go fr the 1) chooise to rename mapFilter -> `mapSkipUndefined` with keep it consistent with current implementation.

When I'll have time I'll rename/rewrite all the files , docs, and tests...

@nartc about testing with `observer-spy` please can you create a commit on this branch, or in main to introduce the new dependency and config for jest or whatever it'll be necessary in the NX workspace to make it works, next I'll rewrite the tests, Thanks

nartc:
We'll revisit the PR after `observer-spy` is in. At the moment, can you fix the rebase since your PR contains code that's already in `main`?

dmorosinotto:
@nartc I'll close this PR and tomorrow will redone all works from ground up in a new PR correct naming `mapSkipUndefined`, and simplifying test using `toArray()` as suggested by @tomalaforge hoping this will make everyone happy and then will be approved.

I need to go sleep tonight, tomorrow I'll done it, promised!

PR #110, opened 2023-10-13
feat(activeElement): introduce injectActiveElement

nevzatopcu:
This PR contains changes that introduce injectActiveElement that returns an Observable of focussed dom element.


libs/ngxtension/active-element/src/lib/active-element.ts
```diff
@@ -0,0 +1,20 @@
+import { DOCUMENT } from '@angular/common';
+import { assertInInjectionContext, inject, type Injector } from '@angular/core';
+import { fromEvent, map, merge, shareReplay } from 'rxjs';
+
+export function injectActiveElement(injector?: Injector) {
+	injector ?? assertInInjectionContext(injectActiveElement);
+	const doc = injector ? injector.get(DOCUMENT) : inject(DOCUMENT);
```
nartc, line 7: suggestion: use `assertInjector` for this. You can look at the other `inject***` to see the usages.

libs/ngxtension/active-element/src/lib/active-element.ts
```diff
@@ -0,0 +1,20 @@
...
+
+	return merge(
+		fromEvent(doc, 'focus', { capture: true, passive: true }).pipe(
+			map(() => true)
+		),
+		fromEvent(doc, 'blur', { capture: true, passive: true }).pipe(
```
nartc, line 13: question: would it make sense to allow the consumers to configure the event options?

nartc (changes requested):
Thank you for your contribution. Couple of things

Also, you should have used `local-plugin:entry-point` to generate the entry point instead of `nx/angular:entry-point`. You can use `local-plugin:convert-entry-point-to-project` generator to convert

nartc:
@all-contributors please add @nevzatopcu for code

libs/ngxtension/active-element/src/lib/active-element.ts
nevzatopcu: EvenListener has 3 options. `capture`, `passive`, and `once`. In this case, 

`capture` must always be set to true; otherwise, it won't work. 
It is advisable to keep `passive` set to true.
Setting `once` to true can disrupt our intended functionality.

Therefore, I would argue that allowing users to edit these event options is unnecessary.

docs/src/content/docs/utilities/Injectors/active-element.md
```diff
@@ -0,0 +1,60 @@
...
+
+The `injectActiveElement` function accepts an optional `Injector` parameter, enabling usage outside of an injection context.
+
+```ts
+@Component()
+export class ExampleComponent implements OnInit {
+	private readonly injector = inject(Injector);
+
+	ngOnInit() {
+		activeElement$ = injectActiveElement(this.injector);
```
eneajaho, line 47:
```suggestion
		const activeElement$ = injectActiveElement(this.injector);
```

libs/ngxtension/active-element/src/active-element.ts
```diff
@@ -0,0 +1,22 @@
+import { DOCUMENT } from '@angular/common';
+import { inject, Injector } from '@angular/core';
+import { assertInjector } from 'ngxtension/assert-injector';
+import { fromEvent, map, merge, shareReplay } from 'rxjs';
+
+export function injectActiveElement(injector?: Injector) {
+	return assertInjector(injectActiveElement, injector, () => {
+		const assertedInjector = inject(Injector);
+		const doc = assertedInjector.get(DOCUMENT);
```
nartc, line 9:
```suggestion
		const doc = inject(DOCUMENT);
```

nartc (approved):

PR #113, opened 2023-10-14
feat(map-skip-undefined): add mapSkipUndefined + filterUndefined operator

dmorosinotto:
This PR substitute #101 to add `mapSkipUndefined` and `filterUndefined` instead of old `mapFilter` proposed in #34
I also added simplified tests using `toArray()` as suggested by @tomalaforge + docs for both operators.
Hope this can be accepted.

### DIFF
--- a/libs/ngxtension/map-skip-undefined/src/map-skip-undefined.spec.ts
+++ b/libs/ngxtension/map-skip-undefined/src/map-skip-undefined.spec.ts
@@ -1,0 +1,58 @@
+import { from, of, toArray } from 'rxjs';
+import { filterUndefined, mapSkipUndefined } from './map-skip-undefined';
+
+describe(filterUndefined.name, () => {
+	it('given an observable of null, undefined, 42, filter out the undefined, emit null and 42', (done) => {
+		const in$ = of(null, undefined, 42);
+		const out$ = in$.pipe(filterUndefined());
+
+		out$.pipe(toArray()).subscribe((r) => {
+			expect(r).toEqual([null, 42]);
+			done();
+		});
+	});
+});
+
+describe(mapSkipUndefined.name, () => {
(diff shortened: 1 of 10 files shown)


dmorosinotto:
@nartc  I don't get why CI is failing, the error is on code **outside** my commit, and before redoing all the work I synced my `main` branch _with yours_ and then created the new branch to easily merge (fast-forward) my new PR at the end...!?

eneajaho:
@nartc lets merge this first and then #115

libs/ngxtension/map-skip-undefined/src/map-skip-undefined.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+				if (n % 2) return String(n * 2);
+				else return undefined; // explict return undefined to skipout (filter) some value from the out observable
+			})
+		);
+
+		out$.pipe(toArray()).subscribe((r) => {
+			expect(r).toEqual(['2', '6']);
+			done();
+		});
+	});
```
tomalaforge, line 30: simple questio: why did you go with 'toArray' instead of observer-spy?

tomalaforge (approved):

libs/ngxtension/map-skip-undefined/src/map-skip-undefined.spec.ts
nartc: @tomalaforge observer-spy commit was reverted due to some ESM issue.

nartc: observer-spy commit was reverted due to some ESM issue

dmorosinotto: @tomalaforge yes I see you commit to introduce observer-spy, and later @nartc revert it
So I was confused, and decided to go to _"easy way"_ rewriting test with `toArray` as you suggested 👍

PR #115, opened 2023-10-14
feat: added inject-lazy

### LINKED ISSUE #44: [feat] add lazyService fn

eneajaho:
This is a new PR as I planned to close the previous one https://github.com/nartc/ngxtension-platform/pull/80

I implemented some stuff with the implementation from @JeanMeche. Also added a test using mockLazyProvider util.

The impl uses observables instead of promises and I changed the name to `injectLazy` instead of `lazyService`

cc @JeanMeche would like you to give it another round of review. 

closes #44

### DIFF
--- a/docs/src/content/docs/utilities/Injectors/inject-lazy.md
+++ b/docs/src/content/docs/utilities/Injectors/inject-lazy.md
@@ -1,0 +1,117 @@
+---
+title: injectLazy
+description: ngxtension/inject-lazy
+---
+
+`injectLazy` is a helper function that allows us to lazily load a service or any kind of Angular provider.
+
+Lazy loading services is useful when we want to shrink the bundle size by loading services only when they are needed.
+
+```ts
+import { injectLazy } from 'ngxtension/inject-lazy';
+```
+
+:::tip[Inside story of the function]
+Initial implementation inspiration: [Lazy loading services in Angular. What?! Yes, we can.](https://itnext.io/lazy-loading-services-in-angular-what-yes-we-can-cfbaf586d54e)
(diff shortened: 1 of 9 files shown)


docs/src/content/docs/utilities/Injectors/inject-lazy.md
```diff
@@ -0,0 +1,117 @@
...
+	template: '<div>{{data$ | async}}</div>',
+})
+class TestComponent {
+	private dataService$ = injectLazy(DataServiceImport);
+
+	data$ = this.dataService$.pipe(switchMap((s) => s.data$));
+}
+```
+
+We can also use `lazyService` not in an injection context, by passing an injector to it.
```
nartc, line 55: nit: `injectLazy` instead of `lazyService` here?

libs/ngxtension/inject-lazy/src/inject-lazy.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @returns An observable of the service.
+ *
+ * @example
+ * ```ts
+ * const dataService$ = lazyService(() => import('./data-service').then((m) => m.MyService));
+ * or
+ * const dataService$ = lazyService(() => import('./data-service'));
+ * ```
+ */
+export function lazyService<T>(
```
nartc, line 19: question: what is `lazyService` for?

eneajaho: This shouldn't be here, removing now.

nartc (approved):

PR #117, opened 2023-10-18
feat: add click outside directive

### LINKED ISSUE #31: [feat] add clickOutside directive

dalenguyen:
This PR adds `clickOutside` directive that fixes #31


nartc:
Looks like there are some lint errors. Can you take a look real quick?

nartc:
ah snap, lint errors are my fault. I'll update main real quick

libs/ngxtension/click-outside/src/click-outside.ts
```diff
@@ -0,0 +1,62 @@
...
+	ngOnInit() {
+		this.documentClick.click$
+			.pipe(takeUntil(this.destroy$))
+			.subscribe((event: Event) => {
+				const isClickedInside = this.elementRef.nativeElement.contains(
+					event.target
+				);
+
+				if (!isClickedInside)
+					this.ngZone.run(() => this.clickOutside.emit(event));
```
ajitzero, line 59:
```suggestion
			.pipe(
			    map(event => this.elementRef.nativeElement.contains(
					event.target
				)),
				filter(isClickedInside => !isClickedInside),
			    takeUntil(this.destroy$),
			)
			.subscribe((event: Event) => {
				this.ngZone.run(() => this.clickOutside.emit(event));
```
Suggestion: I'm not very opinionated about this but I feel moving these conditions to the pipe reads better. What do you think?

libs/ngxtension/click-outside/src/click-outside.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/*
+ * This service is used to detect clicks in the document.
+ * It is used by the clickOutside directive.
+ */
+@Injectable({ providedIn: 'root' })
+export class DocumentClickService {
+	click$ = new Subject<Event>();
+
+	constructor(ngZone: NgZone) {
```
ajitzero, line 23: nit: Can you please use the inject function here instead? This is purely to be consistent with the rest of the files, including this one, where below we can see inject being used.

Can be skipped if it's too much trouble.

libs/ngxtension/click-outside/src/click-outside.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/*
+ * This directive is used to detect clicks outside the element.
+ *
+ * Example:
+ * <div (clickOutside)="close()"></div>
+ *
+ */
+@Directive({ selector: '[clickOutside]', standalone: true })
+export class ClickOutsideDirective implements OnInit {
```
nartc, line 38: nit: I would remove `Directive` from the name of the directive, just `ClickOutside` is enough

libs/ngxtension/click-outside/src/click-outside.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { OnInit } from '@angular/core';
+import { injectDestroy } from 'ngxtension/inject-destroy';
+import { fromEvent, Subject, takeUntil } from 'rxjs';
+
+/*
+ * This service is used to detect clicks in the document.
+ * It is used by the clickOutside directive.
+ */
+@Injectable({ providedIn: 'root' })
+export class DocumentClickService {
```
nartc, line 20: suggestion: this can easily be a CIF instead of a full-blown class, especially we don't even expose this as a public API for `click-outside`

```ts
const [injectDocumentClick] = createInjectionToken(() => {
    const click$ = new Subject<MouseEvent>();
    const [ngZone, document] = [inject(NgZone), inject(DOCUMENT)];
    ngZone.runOutsideAngular(() => {
        fromEvent(document, 'click').subscribe(click$);
    });
    return click$;
})

// then in the directive
export class ClickOutside {
    documentClick$ = injectDocumentClick();
}
```
})

libs/ngxtension/click-outside/src/click-outside.ts
```diff
@@ -0,0 +1,62 @@
...
+	private documentClick = inject(DocumentClickService);
+
+	private destroy$ = injectDestroy();
+
+	/*
+	 * This event is emitted when a click occurs outside the element.
+	 */
+	@Output() clickOutside = new EventEmitter<Event>();
+
+	ngOnInit() {
```
nartc, line 50: question: why does this have to be in `ngOnInit`?

nartc (changes requested):

libs/ngxtension/click-outside/src/click-outside.ts
dalenguyen: Nice suggestion!

libs/ngxtension/click-outside/src/click-outside.ts
dalenguyen: Thanks. I will update based on @nartc suggestion.

libs/ngxtension/click-outside/src/click-outside.ts
dalenguyen: Nice suggestion 🙏

libs/ngxtension/click-outside/src/click-outside.ts
dalenguyen: I think the best lifecycle hook for this task: `Initialize the directive or component after Angular first displays the data-bound properties and sets the directive or component's input properties.` 

https://angular.io/guide/lifecycle-hooks#oninit

ajitzero (approved):

nartc (approved):

nartc:
@all-contributors please add @dalenguyen for code

Issue #37, opened 2023-09-13
RFC add initialValue to computedFrom fn


dmorosinotto:
I follow all your works (article/twit) about `computedForm` and various iterations, and then tried to add something to it: the optional `initialValue` changing a bit of the internal implementation, to not discover the initial values from Obs, but simply relay to the Angular toObservable option using "requireSync" [here](https://gist.github.com/dmorosinotto/0d3d27423e2e8e417ec8f22cc05c1905#file-computedfrom-ts) you can find the code, hope this can be considered or discussed...  

I propose this because I agree that you can easily solve the problem of initial "sync value" for observables using startWith, but I was concerned of cases where the user forgot this and have failing code, adding simple optional to set the initialValue of the resulting Output type maybe and easy task, and at the end if the user just used the startWith in the observable the initial combineLatest have all ready values to compute the first emit, and calling the toObservable with requiredSync will NOT error, but if you have some "later Observable" where you forgot to use startWith and you DON'T pass initialValue it will generate the runtime error, so the user know how to fix the problem (maybe)

nartc:
cc @eneajaho I'll let you discuss this one

eneajaho:
Hello @dmorosinotto 
In fact the current implementation doesn't throw an error, but gives that observable a null initialValue. 
We thought on that (throwing an error or just returning null), and returning null imo was the right choice, because this way the user will be able to see their value will return null. Throwing an error is too intrusive imo. 

[This example from the docs explain that part carefully. 
](https://ngxtension.netlify.app/utilities/computed-from/#3-combine-multiple-signals-and-observables)
```ts
const page$ = new Subject<number>(); // Subject doesn't have an initial value
const filters$ = new BehaviorSubject({ name: 'John' });
const combined = computedFrom([page$, filters$]);

console.log(combined()); // [null, { name: 'John' }] // we don't want console.log to throw an error here
```

I hope I was clear, and if not, please let me know what exactly do you think, by also providing some code we can look at. 

I highly appreciate this discussions.

dmorosinotto:
[...]

PR #122, opened 2023-10-22
fix(computedFrom): add initialValue + throw Error in case of not sync emit

dmorosinotto:
Hi @eneajaho I've finally find the time to make some changes to computedFrom as prompted in #37 

1. My first commit add some more **tests** to your current implementation to prove the issue that I have with _"spurious"_ sync emit of `null` or initial input in case where observable don't have a valid sync emitted value. 

You document those cases as `tricky part` in the docs, but I believe that we can handle it differently throwing an error as the "native" Angular `toSignal` do, so the developer knows where his code will generate the problem and decide how to fix it!

2. I changed the `computedFrom` implementation and tests to handle the `initialValue: Output` passed in options object as last parameter (the options object can be used even to pass the `injector` as before) - I rewrote the test that use the new behavior.

3. In the last commit I updated the `docs` to explain these new behaviors: initialValue + throwing Error.

I hope you can review this PR and give me some feedback, because in the old Issue we didn't have the opportunity to discuss much about how to proceed, and so I decided to complete the code and tests to have a discussion base, and maybe fix or improve it even more to have a  _"better"_ `computedFrom` 🤓


dmorosinotto:
Hi @eneajaho sorry to directly push my code into PR without previous discussion in the issue #37 

But if you agree we can discuss/refine the idea about computedFrom here directly on the code side 😉

I added _“many comments”_ to explain my ideas and motivation, obviusly we’ll remove them befote merge, but let’s start the discussion and give me any feedback about my changes, even what you think about the breaking behaviour of throwing error 👀

dmorosinotto:
Hi @eneajaho , @nartc I know you are both very busy following all the latest Angular news & community, but is there a chance someone would take a look at this PR and gives me any feedback or maybe start a discussion on what you don't like, to make it progress and eventually the changes needed before merging it?

Thanks in advance and sorry to bother you, I really just want to be collaborative... not criticism!

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -71,7 +73,8 @@ let c = computedFrom(
 			([a, b]) =>
 				// of(a + b) is supposed to be an asynchronous operation (e.g. http request)
 				of(a + b).pipe(delay(1000)) // delay the emission of the combined value by 1 second for demonstration purposes
-		)
+		),
+		{ initialValue: 42 } // change the initial value of the resulting signal
```
eneajaho, line 77: We don't need this here as signals have an initial value by default, so it's not the case. 
```suggestion
```

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -87,28 +90,28 @@ setTimeout(() => {
 The console log will be:
 
 ```ts
--[1, 2] - // initial value
+42 - // initial value
```
eneajaho, line 93: Revert this please

docs/src/content/docs/utilities/Signals/computed-from.md
eneajaho, line 77: I'd like to have this here as is. And add another example where we use initial value to showcase how it changes the output.

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -87,28 +90,28 @@ setTimeout(() => {
...
 	3 - // combined value after 1 second
 	5; // combined value after 3 seconds
 ```
 
-As we can see, the first value will not be affected by the rxjs operators, because they are asynchronous and the first value is emitted synchronously.
-In order to change the first value, we can use startWith operator.
+As we can see, we passed an `initialValue` as third argument, to prevent _throwing error_ in case of observable that will **not has sync value** becouse they emit later their values.
```
eneajaho, line 98:
```suggestion
As we can see, we passed an `initialValue` as the third argument, to prevent returning `null` in case of an observable that will **not have a sync value** because they emit later their values.
```

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -151,15 +154,13 @@ console.log(combinedObject()); // { page: 1, filters: { name: 'John' } }
 ```
 
 :::note[Tricky part]
-For `Observable`s that don't emit synchronously, `computedFrom` will give us null as the initial value for the `Observable`s.
+For `Observable`s that don't emit synchronously `computedFrom` will **throw and error** forcing you to fix this situation either passing an `initialValue` in third argument, or using `startWith` operetor to force observable to have a sync starting value.
```
eneajaho, line 157:
```suggestion
For `Observable`s that don't emit synchronously `computedFrom` will **throw an error** forcing you to fix this situation either by passing an `initialValue` in the third argument, or using `startWith` operator to force observable to have a sync starting value.
```

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -151,15 +154,13 @@ console.log(combinedObject()); // { page: 1, filters: { name: 'John' } }
...
 :::
 
 ```ts
 const page$ = new Subject<number>(); // Subject doesn't have an initial value
 const filters$ = new BehaviorSubject({ name: 'John' });
-const combined = computedFrom([page$, filters$]);
-
-console.log(combined()); // [null, { name: 'John' }]
+const combined = computedFrom([page$, filters$]); // 👈 WILL THROW AN ERROR!!!
```
eneajaho, line 163:
```suggestion
const combined = computedFrom([page$, filters$]); // 👈 will throw an error!! 💥
```

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -173,9 +174,20 @@ const combined = computedFrom([
 console.log(combined()); // [0, { name: 'John' }]
 ```
 
-### 4. Use it outside of an injection context
+### 4. Use of options initialValue
```
eneajaho, line 177:
```suggestion
### 4. Using initialValue param
```

libs/ngxtension/computed-from/src/computed-from.spec.ts
```diff
@@ -25,6 +25,7 @@ import {
 	switchMap,
 } from 'rxjs';
 import { computedFrom } from './computed-from';
+import exp = require('constants');
```
eneajaho, line 28: Is this needed?

libs/ngxtension/computed-from/src/computed-from.spec.ts
```diff
@@ -58,6 +59,30 @@ describe(computedFrom.name, () => {
 				expect(s()).toEqual([1]);
 			});
 		});
+		it(`MD FIX for Observables that don't emit synchronously, computedFrom will THROW ERROR`, () => {
+			TestBed.runInInjectionContext(() => {
+				const late = of(1).pipe(delay(1000)); // late emit after 1s
+				expect(() => {
+					const s = computedFrom([late]);
+				}).toThrowError(/requireSync/i); //THROW ERROR NG0601 DUE TO toSignal + requireSync: true
```
eneajaho, line 67:
```suggestion
				}).toThrowError(/requireSync/i); // Throw error NG0601 due to `toSignal` + `requireSync: true`
```

libs/ngxtension/computed-from/src/computed-from.spec.ts
```diff
@@ -58,6 +59,30 @@ describe(computedFrom.name, () => {
...
+				//THIS WILL PREVENT OLD "SPURIOUS SYNC EMIT" OF null OR Input ([], {}) THAT CAN CAUSE TS RUNTIME ERRORS
```
eneajaho, line 68: Can you use normal lowercase sentences, please? 

Uppercase is not easy to read.

libs/ngxtension/computed-from/src/computed-from.spec.ts
```diff
@@ -58,6 +59,30 @@ describe(computedFrom.name, () => {
...
+				// expect(() => s()[0].toFixed(2)).toThrowError(/null/i); //NOTICE THAT THIS WILL EXPLODE AT RUNTIME - TS DON'T CATCH IT!!!
+				// tick(1000); //WAIT 1s FOR LATE EMIT
+				// expect(s()).toEqual([1]); //NOW WE HAVE THE REAL VALUE
+				// expect(s()[0].toFixed(2)).toEqual('1.00'); //HERE WE CAN CALL s()[0].toFixed(2) <-- THIS WILL WORK
+			});
+		});
+		it(`MD FIX for Observables that don't emit synchronously, you can pass options.initialValue TO PREVENT ERROR`, fakeAsync(() => {
```
eneajaho, line 75: What is MD ?

libs/ngxtension/computed-from/src/computed-from.ts
```diff
@@ -59,53 +86,82 @@ export function computedFrom<Input extends object, Output = Input>(
...
+): Signal<Output> {
+	const { normalizedSources, hasInitValue, operator, options } = _normalizeArgs<
+		Input,
+		Output
+	>(args);
+
+	let injector = options?.injector;
 	injector = assertInjector(computedFrom, injector);
+	/* try { //CUSTOM ERROR HANDLING FOR computedFrom */
+	//IF YOU PASS options.initialValue RETURN Signal<Output> WITHOUT ANY PROBLEM EVEN IF sources Observable ARE ASYNC (LATE EMIT) -> OUTPUT SIGNAL START EMITING SYNC initialValue!
```
eneajaho, line 100: Can you convert the text to lowercase, please?

libs/ngxtension/computed-from/src/computed-from.ts
```diff
@@ -59,53 +86,82 @@ export function computedFrom<Input extends object, Output = Input>(
...
+		  });
+	return ret;
+	/* //WE CAN DECIDE TO CUSTOMIZE THE ERROR TO BE MORE SPECIFIC FOR computedFrom
+	} catch (e: any) {
+		if ( //EURISTIC TO CHECK IF THE ERROR IS CAUSED BY requireSync
+			e.message.includes('requireSync') ||
+			e.message.includes('NG601') ||
+			e.code == 601
+		)
+			console.warn(
```
eneajaho, line 121: @nartc What do you think about this one? 

I think we will need sth like this.

libs/ngxtension/computed-from/src/computed-from.ts
```diff
@@ -59,53 +86,82 @@ export function computedFrom<Input extends object, Output = Input>(
...
+	args: any[]
+): {
+	normalizedSources: ObservableInputTuple<Input>;
+	operator: OperatorFunction<Input, Output>;
+	hasInitValue: boolean;
+	options: ComputedFromOptions<Output> | undefined;
+} {
+	if (!args || !args.length || typeof args[0] !== 'object')
+		//VALID EVEN FOR ARRAY
+		throw new TypeError('computedFrom need sources');
```
eneajaho, line 142:
```suggestion
		throw new TypeError('computedFrom needs sources');
```

libs/ngxtension/computed-from/src/computed-from.ts
```diff
@@ -59,53 +86,82 @@ export function computedFrom<Input extends object, Output = Input>(
...
+	const hasOperator = typeof args[1] === 'function';
+	if (args.length == 3 && !hasOperator)
+		throw new TypeError('computedFrom need pipebale operator as second arg');
```
eneajaho, line 145:
```suggestion
		throw new TypeError('computedFrom needs pipeable operator as a second argument');
```

eneajaho:
Hi @dmorosinotto 

It looks good to me. It doesn't look breaking to me. There are some comments I'd like you to review. 

And we need to discuss the error-handling part with @nartc . 

Thanks for your patience.

dmorosinotto:
Hi @eneajaho thanks for your feedback reviewing my code, today I'm quite busy at work, but I'll take all your suggestions to update my code this evening (after 21:00 CEST) 
Just to give you some quick response (than I'll detail it in context of your comments)
- I'll remove/lower all UPPERCASE comments
- about the reverted initialValue sample where in your original implementation we get the [1,2] emission, it's not possible becouse in case of "async later emit" the new code will `throwError` becouse I added `pipe(delay(1000))` 
- about the `MD & MD FIX` in the tests title it was just my signature to easly find/quick identify where I changed the behavior / breacking change respect to your older behavior, I'll *absolutely remove* `MD` from the final code!
-  I don't know where I added the `exp = required("...")` probably some wrong auto-import, obviously I'll remove it

docs/src/content/docs/utilities/Signals/computed-from.md
dmorosinotto: If I remove the { initialValue: 42 } this sample will THROW becouse it's exactly the case where the input + delay operator generate "late async" emit , so with the new implementation, you don't get the "old spurious" initial emit of array [1,2] (that is what I really don't like because it's totally un-expected (and maybe error prone) with an inferred output as Signal<number>.

This is why I introduced { initialValue: 42 } in the sample, and adjusted the output expected in the console.

Let's me know if you agree, or if you prefer to have a different sample showing the use-case using only inputs + pipe Operator, and keeping this one to show the initialValue to prevent throwing error... 👀

libs/ngxtension/computed-from/src/computed-from.spec.ts
dmorosinotto: No it was an auto-import mistake will remove it

libs/ngxtension/computed-from/src/computed-from.spec.ts
dmorosinotto: Will remove all of it

dmorosinotto:
OK @eneajaho I try to fix the typos and uppercase tests comments, hope this will made it more readable and conform to the repo standards.

Waiting for your feedback and @nartc opinion about the new `throwing error` behavior 👀

docs/src/content/docs/utilities/Signals/computed-from.md
eneajaho: Let's have both cases. The one that throws and the new one that doesn't.

dmorosinotto: Ok @eneajaho I added the first sample without params so it'll throw error, and then show the two possible solutions with initialValue or with startWith. 
Let's me know what you think?

libs/ngxtension/computed-from/src/computed-from.ts
dmorosinotto: Sorry forgot of this uppercase in the code - now I fixed it.
 @nartc & @eneajaho please let's me know if what you think of the new behavior I introduced?

dmorosinotto:
@eneajaho , @nartc I fixed the CI errors - if you want to review the code and give me some feedback about what you think of the current "throwing error" behavior?  

BTW: I had to do some "try&error" to make a commit passing CI, but running local test always pass even with previous failing CI commits, I suspect that the CI is using some stricter linting rules! Can we align those so anyone can verify all code & test works as expected from CI before commit? Thanks

PS: hope to meet you IRL at NGPoland next week, to discuss this or other Angular stuff 😉

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -84,31 +86,46 @@ setTimeout(() => {
 // You can copy the above example inside an Angular constructor and see the result in the console.
 ```
 
-The console log will be:
+This will _throw an error_ because the operation pipeline used will produce an observable that will **not have a sync value** because they emit later their values, so the resultig `c` signal doesn't have an initial value, and this caused the error.
```
eneajaho, line 89:
```suggestion
This will _throw an error_ because the operation pipeline will produce an observable that will **not have a sync value** because they emit their values later on, so the resulting `c` signal doesn't have an initial value, and this causes the error.
```

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -84,31 +86,46 @@ setTimeout(() => {
...
+	pipe(
+		switchMap(
+			([a, b]) => of(a + b).pipe(delay(1000)) // later async emit value
+		),
+		{ initialValue: 42 } // 👈 pass the initial value of the resulting signal
+	)
+);
+```
+
+This will works, and you can copy the above example inside an Angular constructor and see the result in the console:
```
eneajaho, line 105:
```suggestion
This works, and you can copy the above example inside an Angular constructor and see the result in the console:
```

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -84,31 +86,46 @@ setTimeout(() => {
...
+
+```ts
+42 - // initial value passed as third argument
 	3 - // combined value after 1 second
 	5; // combined value after 3 seconds
 ```
 
-As we can see, the first value will not be affected by the rxjs operators, because they are asynchronous and the first value is emitted synchronously.
-In order to change the first value, we can use startWith operator.
+Another way to solve the same problem, is using the `startWith` operator in the pipe to force the observable to have a starting value like below.
```
eneajaho, line 113:
```suggestion
Another way to solve this problem is using the `startWith` rxjs operator in the pipe to force the observable to have a starting value like below.
```

docs/src/content/docs/utilities/Signals/computed-from.md
```diff
@@ -84,31 +86,46 @@ setTimeout(() => {
...
 
 ```ts
 let c = computedFrom(
 	[a, b],
 	pipe(
 		switchMap(([a, b]) => of(a + b).pipe(delay(1000))),
-		startWith(0) // change the first value
+		startWith(0) // 👈 change the starting value (sync emit)
```
eneajaho, line 120:
```suggestion
		startWith(0) // 👈 change the starting value (emits synchronously)
```

eneajaho:
Hi @dmorosinotto 
Thanks, it looks good now. 

Regarding try catch error handling, would you mind opening an issue, so we can tackle it later and get this PR merged? 

Also, please can you fix those other change requests? And after that it should be good to be merged.

libs/ngxtension/computed-from/src/computed-from.ts
dmorosinotto: @eneajaho when you ask me to open a new issue to discuss try catch handle error are you referring to those commented lines where I suggest a way to handle the error due to `requireSync` to customize the message for `computedForm`?

dmorosinotto:
Hi @eneajaho I've added a commit to fix remaining typo in docs.
 
And just for clarification about the issue that you want me to open - are you referring to commented lines in `computedFrom.ts` L113-128 where I suggest a way to handle&customize error message due to `requiredSync`?

nartc (approved):

nartc:
I hate you guys for having fun at ngpoland conf without me so I rage merged this PR

eneajaho:
Oh, I'm so sorry for you @nartc  🫶

PR #135, opened 2023-11-07
feat(signal-slice): add signalSlice

joshuamorony:
This was created in collaboration with Chau, it adds `signalSlice` which is loosely inspired by the `createSlice` API from Redux Toolkit. The general idea is that it allows you to declaratively create a "slice" of state. This state will be available as a readonly signal.

The key motivation, and what makes this declarative, is that all the ways for updating this signal are declared upfront with `sources` and `reducers`. It is not possible to imperatively update the state.

This PR also makes minor changes to `connect` in order to expose types that are shared by this utility.

Example using all API features:

```ts
  state = signalSlice({
    sources: [this.sources$],
    reducers: {
      add: (state, checklist: AddChecklist) => ({
        checklists: [...state.checklists, this.addIdToChecklist(checklist)],
      }),
      edit: (state, update: EditChecklist) => ({
        checklists: state.checklists.map((checklist) =>
          checklist.id === update.id
            ? { ...checklist, title: update.data.title }
            : checklist
        ),
      }),
      remove: (state, id: RemoveChecklist) => ({
        checklists: state.checklists.filter((checklist) => checklist.id !== id),
      }),
    },
    selectors: (state) => ({
      loadedAndError: () => state().loaded && state().error,
      whatever: () => 'hi',
    }),
  });
```

Key features:
* Sources will automatically update signal when they emit
* Reducers will automatically create actions on the state object that can be called (e.g. `this.state.remove(id)`)
* Top-level properties of the initial state are automatically exposed as selectors (computeds) on the state object (e.g. this.state.loaded()`)
* Additional selectors can be supplied which will also be added to the state object

For my own purposes this API is "feature complete" and does everything I need but I'm very open to comments/revisions/improvements.

### DIFF
--- a/docs/src/content/docs/utilities/Signals/signal-slice.md
+++ b/docs/src/content/docs/utilities/Signals/signal-slice.md
@@ -1,0 +1,117 @@
+---
+title: signalSlice
+description: ngxtension/signalSlice
+---
+
+`signalSlice` is loosely inspired by the `createSlice` API from Redux Toolkit. The general idea is that it allows you to declaratively create a "slice" of state. This state will be available as a **readonly** signal.
+
+The key motivation, and what makes this declarative, is that all the ways for updating this signal are declared upfront with `sources` and `reducers`. It is not possible to imperatively update the state.
+
+## Basic Usage
+
+```ts
+import { signalSlice } from 'ngxtension/signal-slice';
+```
+
+```ts
+  private initialState: ChecklistsState = {
+    checklists: [],
+    loaded: false,
+    error: null,
+  };
+
+  state = signalSlice({
+    initialState: this.initialState,
+  });
+```
+
+The returned `state` object will be a standard **readonly** signal, but it will also have properties attached to it that will be discussed below.
+
+You can access the state as you would with a typical signal:
+
+```ts
+this.state().loaded;
+```
+
+However, by default `computed` selectors will be created for each top-level property in the initial state:
+
+```ts
+this.state.loaded();
+```
+
+## Sources
+
+One way to update state is through the use of `sources`. These are intended to be used for "auto sources" — as in, observable streams that will emit automatically like an `http.get()`. Although it will work with a `Subject` that you `next` as well, it is recommended that you use a **reducer** for these imperative style state updates.
+
+You can supply a source like this:
+
+```ts
+this.checklistsLoaded$.pipe(map((checklists) => ({ checklists, loaded: true })));
+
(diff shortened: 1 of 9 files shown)


docs/src/content/docs/utilities/Signals/signal-slice.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+## Sources
+
+One way to update state is through the use of `sources`. These are intended to be used for "auto sources" — as in, observable streams that will emit automatically like an `http.get()`. Although it will work with a `Subject` that you `next` as well, it is recommended that you use a **reducer** for these imperative style state updates.
+
+You can supply a source like this:
+
+```ts
+this.checklistsLoaded$.pipe(map((checklists) => ({ checklists, loaded: true })));
```
nartc, line 49: nit: this should probably be assigned to something else then use that in the `sources` array below?

docs/src/content/docs/utilities/Signals/signal-slice.md
```diff
@@ -0,0 +1,117 @@
...
+});
+```
+
+The `source` should be mapped to a partial of the `initialState`. In the example above, when the source emits it will update both the `checklists` and the `loaded` properties in the state signal.
+
+## Reducers and Actions
+
+Another way to update the state is through `reducers` and `actions`. This is good for situations where you need to manually/imperatively trigger some action, and then use the current state in some way in order to calculate the new state.
+
+When you supply a `reducer` it will automatically create an `action` that you can call. Reducers can be created like this:
```
nartc, line 63:
```suggestion
When you supply a `reducer`, it will automatically create an `action` that you can call. Reducers can be created like this:
```

docs/src/content/docs/utilities/Signals/signal-slice.md
```diff
@@ -0,0 +1,117 @@
...
+	initialState: this.initialState,
+	reducers: {
+		add: (state, checklist: AddChecklist) => ({
+			checklists: [...state.checklists, checklist],
+		}),
+		remove: (state, id: RemoveChecklist) => ({
+			checklists: state.checklists.filter((checklist) => checklist.id !== id),
+		}),
+	},
+});
```
nartc, line 76: suggestion: I'd love to mention a reducer without a payload can simply be called without parameters

nartc (approved):
Some nits but approved

nartc:
@all-contributors please add @joshuamorony for code

PR #144, opened 2023-11-11
feat(signal-slice): add asyncReducers

joshuamorony:
This is a draft implementation that allows optionally supplying sources as a function that can utilise the current state and return an array of sources.

This addresses what I think is a fundamental problem with the current implementation and is highlighted by this use case from @justinrassier on Twitter: https://twitter.com/justinrassier/status/1722619039787446438 - I've also already run into this issue for error handling

Actions can be created by defining reducers, but if you want a source to be able to utilise an action you have to define one separately outside of `signalSlice` (e.g. as in the example above where a separate `load$` action is created with a subject).

This change would allow (optionally) supplying a function instead of an array of sources directly:

```ts
  state = signalSlice({
    sources: (state) => [
      state.manualLoad$.pipe(switchMap(() => this.sources$)),
    ],
    reducers: {
      manualLoad: () => ({}),
    },
  });
```

This allows creating the action inside of `signalSlice` but still also allows it to be utilised by `sources`. This particular example waits for the `manualLoad` action to be triggered before actually loading from the sources.

The current implementation in this PR works, but the typing is broken. I am opening this draft PR to investigate whether it is possible to type this API or whether we can arrive at an alternative solution.

### DIFF
--- a/docs/src/content/docs/utilities/Signals/signal-slice.md
+++ b/docs/src/content/docs/utilities/Signals/signal-slice.md
@@ -101,6 +101,30 @@ The associated action can then be triggered with:
 this.state.toggleActive();
 ```
 
+## Async Reducers
+
+A standard reducer accepts a function that updates the state synchronously. It
+is also possible to specify `asyncReducers` that return an observable to update
+the state asynchronously.
+
+For example:
+
+```ts
+  state = signalSlice({
+    initialState: this.initialState,
+    asyncReducers: {
+      load: ($: Observable<void>) => $.pipe(
+        switchMap(() => this.someService.load()),
+        map(data => ({ someProperty: data })
+      )
+    }
+  })
+```
+
+In this particular case, a `load` action will be created that can be called with `this.state.load()`. When this action is called the internal `Subject` will be nexted, and it is this subject that is being supplied as the `$` parameter above. You can then `pipe` onto that `$` to perform whatever asynchronous operations you need, and then at the end you should `map` the result to whatever parts of the state signal you want to update (just like with standard `reducers`).
+
+**NOTE:** This example covers the use case where data _needs_ to be manually triggered with a `load()` action. It is also possible to just have your data load automatically — in this case the observable that loads the data can just be supplied directly through `sources` and it will be loaded automatically without needing to trigger the `load()` action.
+
 ## Action Streams
 
 The source/stream for each action is also exposed on the state object. That means that you can access:
--- a/docs/src/content/docs/utilities/Signals/signal-slice.md
+++ b/docs/src/content/docs/utilities/Signals/signal-slice.md
@@ -113,7 +113,7 @@ For example:
   state = signalSlice({
     initialState: this.initialState,
     asyncReducers: {
-      load: ($: Observable<void>) => $.pipe(
+      load: (_state, $: Observable<void>) => $.pipe(
         switchMap(() => this.someService.load()),
         map(data => ({ someProperty: data })
       )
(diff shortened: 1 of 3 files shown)


nartc:
The typing issue is pretty tricky. TypeScript cannot prioritize which property to infer `TReducers` from because `sources` and `reducers` both "utilize" the `TReducers` type parameter.

```ts
// this won't work because TReducers from "sources" takes precedence
signalSlice({
  sources: () => [],
  reducers: {}
});

// this WILL work
signalSlice({
  reducers: {},
  sources: () => []
})
```

With that in mind, an API that depends on the order of object properties isn't scalable and hard to reason about. Hence, we shouldn't change how `sources` work, at least for now.

That said, here are 2 alternatives:

- Allow `reducers` to accept **external** `Subject`. This allows for a little less boilerplate on the consumers' side.

```ts
const load$ = new Subject<void>();
const state = signalSlice({
  initialState,
  sources: [load$.pipe(switchMap())],
  reducers: {
    load: load$
  }
});

state.load(); // "nexting" the Subject
```

- Allow `reducers` to accept a Function that returns `Observable<PartialOrValue<TSignalValue>>`. I think this alternative has the best set of trade-off. 

```ts
const state = signalSlice({
  initialState,
  reducers: {
    //    👇 we create this internal "stream" anyway
    //                   👇 this provides the type for "load()" reducer
    load: ($: Observable<void>) => $.pipe(switchMap(/*...*/))
  }
});

state.load(); // this will kick off the stream
```

Let's attempt Justin's example with the 2nd alternative

```ts
type ReferenceState = {
  referenceAsyncState: ReferenceAsyncState;
}

const initialReferenceState: ReferenceState = {
  referenceAsyncState: initialAsyncState()
}

@Injectable({ providedIn: 'root' })
export class ReferenceFacade {
  #referenceService = inject(ReferenceService);
  
  public state = signalSlice({
    initialState: initialReferenceState,
    reducers: {
      load: ($: Observable<void>) => $.pipe(
        switchMap(() => this.#referenceService.load()),
        map(references => ({ referenceAsyncState: loadedAsyncState({ references }) })
      )
    }
  })
}
```

joshuamorony:
I agree, the second approach appeals more to me for two main reasons:

1) Keeps it all inside `signalSlice` - it doesn't feel quite right to create this "state management mechanism" manually (the load subject) outside of `signalSlice` and pass it to `signalSlice`

2) The explanation/story/docs will just make more sense imo, e.g. for the load case:

Option 1: "You can create a load$ subject externally that you pass in as a reducer and also as a source, you can then call the action created by that reducer to next your load$ subject, and since it is also being used as a source it can trigger the load when the action is called"

vs

Option 2: "You can create reducers that can act as manually triggered sources. So, you can create a "load" reducer that returns your source stream, and that source will be triggered when you call state.load()"

imo option 2 is far more intuitive (as well as resulting in a cleaner API). I'll work on this.

nartc:
I have made an adjustment to the API so that the reducer always receives the `state` as its first argument.
```ts
const state = signalSlice({
  initialState,
  reducers: {
    load: (state, $: Observable<void>) => $.pipe(switchMap(/*...*/))
  }
});
```

docs/src/content/docs/utilities/Signals/signal-slice.md
```diff
@@ -101,6 +101,30 @@ The associated action can then be triggered with:
...
+is also possible to specify `asyncReducers` that return an observable to update
+the state asynchronously.
+
+For example:
+
+```ts
+  state = signalSlice({
+    initialState: this.initialState,
+    asyncReducers: {
+      load: ($: Observable<void>) => $.pipe(
```
nartc, line 116:
```suggestion
      load: (_state, $: Observable<void>) => $.pipe(
```

nartc (approved):

PR #154, opened 2023-11-16
feat(signal-slice): add actionEffects config

joshuamorony:
EDIT: The implementation of this has changed quite a bit since I did it initially so I've updated this comment to reflect that.

There might still be a bit of tweaking to do here (I haven't been able to properly type action.payload but all the other properties are fine), and I will need to write some docs, but it's at a state where it would be great to get some feedback. It works as is, I'm just not 100% sure on the API.

The initial motivation for this was an example from @nartc attempting to implement Real World app and running into an issue where an `actionSource` would need to run and once that completed a side effect (navigation) would need to occur with a dynamic route.

Currently with signalSlice, the primary way to execute effects is by reacting to a state change, but it seems useful and necessary to run side effects for streams too.

This PR contains an implementation to add `actionEffects`, generally they work like this (the actionEffect will be triggered when the source for the relevant action emits):

```ts
  state = signalSlice({
    sources: [this.sources$],
    actionSources: {
      login: (state, action$: Observable<string>) =>
        action$.pipe(
          switchMap(() => this.getTemporalUser()),
          map(() => ({}))
        ),
    },
    actionEffects: () => ({
      login: (action) => {
        console.log(action.name); // the name of the action triggered
        console.log(action.payload); // the value the action was called with
        console.log(action.value); // the value the source emitted with (if successful, undefined otherwise)
        console.log(action.err); // the error the source emitted (if errored, undefined otherwise)
      },
    }),
  });
```

This should solve the use case faced in Real World app:

```ts
  state = signalSlice({
    sources: [this.sources$],
    actionSources: {
      refresh: (state, action$: Observable<string[]>) =>
        action$.pipe(
          switchMap(() => doWhatever()),
          map(() => ({}))
        ),
    },
    actionEffects: () => ({
      refresh: (action) => {
        // navigate using params from action.payload
      },
    }),
  });
```

`actionEffects` are also passed the signal slice, so you could trigger more actions within actionEffects:

```ts
    actionEffects: (state) => ({
      someAction: (action) => {
        state.someOtherAction();
      },
    }),
```

**Other considerations**

* I've included `action.name` in the action data as I am intending to investigate adding support for an "all actions" stream, e.g. an action stream that will be triggered when any action is triggered. The plan is to allow creating an actionEffect that could react to this all actions stream, and being able to identify the action by name would be used - If I end up doing this it will be in a separate PR
* The standard `effects` config still exists which uses standard signal effects to react to signal state changes. I'm not sure if this is confusing or if its even really necessary to have this config option at all. Renaming `effects` to `signalEffects` or `stateEffects` might also be an option, or we could just leave it as is - not sure

### DIFF
--- a/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
+++ b/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
@@ -305,6 +305,30 @@ describe(signalSlice.name, () => {
 		});
 	});
 
+	describe('actionEffects', () => {
+		it('should create effects for named actionEeffects', () => {
+			TestBed.runInInjectionContext(() => {
+				const testFn = jest.fn();
+
+				const state = signalSlice({
+					initialState,
+					reducers: {
+						increaseAge: (state) => ({ age: state.age + 1 }),
+					},
+					actionEffects: (state) => ({
+						increaseAge$: (val) => {
+							testFn(val.age);
+						},
+					}),
+				});
+
+				state.increaseAge();
+
+				expect(testFn).toHaveBeenCalledWith(initialState.age + 1);
+			});
+		});
+	});
+
 	describe('effects', () => {
 		it('should create effects for named effects', () => {
 			TestBed.runInInjectionContext(() => {
--- a/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
+++ b/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
@@ -358,6 +358,84 @@ describe(signalSlice.name, () => {
 				expect(testFn).toHaveBeenCalled();
 			});
 		});
+
+		it('should supply appropriate values on action', () => {
+			TestBed.runInInjectionContext(() => {
+				const testFn = jest.fn();
+				const testPayload = 'a';
+				const age = 20;
+
+				const state = signalSlice({
+					initialState,
+					actionSources: {
+						test: (state, $: Observable<string>) =>
+							$.pipe(
+								map(() => ({
+									age,
+								}))
+							),
+					},
+					actionEffects: () => ({
+						test: (action) => {
+							testFn({
+								name: action.name,
+								payload: action.payload,
+								value: action.value,
+								err: action.err,
+							});
+						},
+					}),
+				});
+
+				state.test(testPayload);
+
+				expect(testFn).toHaveBeenCalledWith({
+					name: 'test',
+					payload: testPayload,
+					value: { age },
+					err: undefined,
+				});
+			});
+		});
+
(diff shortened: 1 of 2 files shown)


joshuamorony:
Another potential way to deal with confusion as to the role of effects here is to not have "state" effects - the current implementation is no different than just defining an external signal effect, it just co-locates it with the state definition. But at least so far I have found I want to run my signal effects in other places as well anyway (e.g. my `signalSlice` is in the service, but I am setting up signal effects in components)

So, maybe if you want to define a state/signal effect you just do that externally. Then `effects` in signalSlice can become what I am suggesting for `actionEffects` here (i.e. side effects that are triggered in response to the actions).

critchie:
Currently I'm doing something like this (I have not converted to signalSlice yet):
```
this.save$
  .pipe(
    takeUntilDestroyed(),
    map(() => {
      // build new location from form values
      return this.locationFormGroup().value;
    }),
    switchMap((location: Location) =>
      this._locationsApiService.updateLocation(location)
    )
  )
  .subscribe((location: Location) => {
    this._state.update((state) => ({
      ...state,
      location,
    }));
    this._titleService.setTitle(`Locations - ${this.location()?.name}`);
    this._applicationState.showMessage$.next({
      severity: 'success',
      summary: 'Location Details',
      detail: 'Saved location successfully.',
    });
    this.canNavigate$.next(true);
  });
```
It is these lines, the UX things that need to be done after a successful call to the API, that don't feel right in the subscribe block:
```
this._titleService.setTitle(`Locations - ${this.location()?.name}`);
this._applicationState.showMessage$.next({
  severity: 'success',
  summary: 'Location Details',
  detail: 'Saved location successfully.',
});
this.canNavigate$.next(true);
```

Is this the justification for the actionEffects? The place to put things after the state is updated?

joshuamorony:
Of course these are ultimately my ideas/preferences around declarative code and not something you *have* to do, but for the vanilla approach generally the idea would be to never have anything but the `state.update` call in the subscribe, any additional code in there is just adding more imperative code.

For this particular approach you likely wouldn't need actionEffects - rather than calling the "setTitle" method I would have the title derived declaratively based on the location in the state, the "showMessage" toast can be triggered using a normal signal effect that reacts to the location state changing in the signal, and as for the "canNavigate" flag I would need more context but this is also probably an extra imperative step that wouldn't be required.

actionEffects is really for things that have to be side effects, but currently the only way with this approach to create side effects is by reacting to a change in the state signal. Sometimes you might need a side effect that doesn't result from a change in the state, and this is where actionEffects would become useful (running side effects as a result of stream emissions rather than signal state changes essentially)

Thanks for the comment/thoughts!

nartc (approved):

PR #162, opened 2023-11-19
fix(create-injection-token): better decide when to use factory or provided value (#161)

### LINKED ISSUE #161: createInjectionToken wrongfully disregards falsy argument passed to generated provider function
Suppose for whatever reason you want to provide a value like this:

```ts
const [injectSomeValue, provideSomeValue] = createInjectionToken(factory);

function factory(): number {
  return 1;
}
```

Later on you actually provide a value so it can be injected instead of the factory generated one:
```ts
providers: [provideSomeValue(0)]
```

Because of how the provider function (`provideSomeValue`) created by [createProvideFn](https://github.com/nartc/ngxtension-platform/blob/7df93d303cb24022dfb25602163ec46d9a4a2972/libs/ngxtension/create-injection-token/src/create-injection-token.ts#L116C1-L137C15) decides whether to use the configured factory or the value that was provided to `provideSomeValue`, any provided value that evaluates to falsy (0, "", false, null, undefined, etc) will be disregarded and the factory will be used instead.

Instead of evaluating the optionally passed argument itself, maybe evaluate if an argument was actually passed instead?

diegovilar:
This PR fixes #161

### DIFF
--- a/libs/ngxtension/create-injection-token/src/create-injection-token.ts
+++ b/libs/ngxtension/create-injection-token/src/create-injection-token.ts
@@ -113,7 +113,7 @@ function createProvideFn<
 	const { deps = [], multi = false, extraProviders = [] } = opts;
 	return (value?: TValue | (() => TValue), isFunctionValue = false) => {
 		let provider: Provider;
-		if (value) {
+		if (typeof value !== 'undefined') {
 			// TODO: (chau) maybe this can be made better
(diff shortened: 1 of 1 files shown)


nartc (approved):

nartc:
@all-contributors please add @diegovilar for code

nartc (approved):

PR #174, opened 2023-11-23
feat: add collection for contributors

tomalaforge:
To follow up on the contributor button and from the idea of @ajitzero, I added a new collection to collect contributors data where we can add a link to twitter, github, linkedin. Other link can be added.

### DIFF
--- a/docs/src/content/config.ts
+++ b/docs/src/content/config.ts
@@ -1,13 +1,28 @@
 import { docsSchema, i18nSchema } from '@astrojs/starlight/schema';
-import { defineCollection, z } from 'astro:content';
+import { defineCollection, reference, z } from 'astro:content';
+
+const contributors = defineCollection({
+	type: 'data',
+	schema: z.object({
+		name: z.string(),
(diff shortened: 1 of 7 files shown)


docs/src/content/config.ts
```diff
@@ -1,13 +1,28 @@
...
+			contributor: reference('contributors').optional(),
+		}),
+});
+
+const i18n = defineCollection({ type: 'data', schema: i18nSchema() });
+
+export const collections = {
+	docs: docs,
+	i18n: i18n,
+	contributors: contributors,
```
ajitzero, line 27: nit: I prefer the [shorthand](https://javascript.info/object#property-value-shorthand), mainly due to [this ES Lint rule](https://eslint.org/docs/latest/rules/object-shorthand). I have no strong opinions on this though.

```suggestion
	docs,
	i18n,
	contributors,
```

ajitzero (approved):
That was fast! I've left a minor comment/question regarding storing just the usernames instead of the full links, but either way is fine!

Issue #176, opened 2023-11-24
[Suggestion] Do not allow optional properties in signalSlice


palexcast:
Optional properties in the initial state pose an issue with selectors.

Their type is `Signal<value | undefined> | undefined` which makes sense, as we need all they keys of the initialState to be there to create a signal for it.

I would expect it to be `Signal<value | undefined` but as stated above, this is impossible without the key.

### Proposal
To fix this, you could disallow optional properties in the signalSlice. 
I created a PR that I think fixes it, but the downside is that it might be confusing for someone who doesn't understand the difference between `?:` and `| undefined`.

PR: https://github.com/nartc/ngxtension-platform/pull/177

#### With `?:`
```
  type InitialStateWithUndefined = { thisShouldExist: string | undefined };
  let state: SignalSlice<InitialStateWithUndefined, any, any, any> = signalSlice({
    initialState: {
      thisShouldExist: undefined,
    }
  });
  state.thisShouldExist(); // type is
```

screenshot from Webstorm:

![image](https://github.com/nartc/ngxtension-platform/assets/15246162/6d2ddae9-9f06-4a18-98e2-c47df8b13b86)

#### With `undefined` instead of `?:`
```
  type InitialStateWithUndefined = { thisShouldExist: string | undefined };
  let state: SignalSlice<InitialStateWithUndefined, any, any, any> = signalSlice({
    initialState: {
      thisShouldExist: undefined,
    }
  });
  state.thisShouldExist(); // type is `thisShouldExist?: Signal<string | undefined> | undefined`
```

screenshot from webstorm:
![image](https://github.com/nartc/ngxtension-platform/assets/15246162/a0d94094-d11d-4dc7-83e0-aca9c2642fdd)

palexcast:
Was solved in https://github.com/nartc/ngxtension-platform/pull/177

PR #177, opened 2023-11-24
fix(signal-slice): Do not allow optional properties in signalSlice

palexcast:
Optional properties in the initial state pose an issue with selectors, this PR forces a user to set it as undefined instead.

### With `?:`
image

### With `| undefined`
image

Issue explaining in detail: https://github.com/nartc/ngxtension-platform/issues/176

### DIFF
--- a/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
+++ b/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
@@ -33,7 +33,7 @@ describe(signalSlice.name, () => {
 			expect(state.age()).toEqual(initialState.age);
 		});
 
-		it('optional properties are not allowed in initial state', () => {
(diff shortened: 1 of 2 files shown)


joshuamorony:
Thanks for this @palexcast - I'm happy with the change, and I don't know how else it would be fixed anyway (although @nartc has dealt with most of the tricker typing situations for signalSlice so he might have thoughts). I think we can just add a brief note about this to the docs under Basic Usage to explain that it is not allowed (we probably don't need to get into the specifics right away).

nartc:
@all-contributors please add @palexcast for code

nartc (approved):

nartc:
LGTM. Can you add some comment to the ts-expect-error directive?

palexcast:
> LGTM. Can you add some comment to the ts-expect-error directive?

Of course, done 👍

PR #179, opened 2023-11-28
chore: update enea jahollari info and links

eneajaho:


docs/src/components/Contributor.astro
```diff
@@ -4,11 +4,12 @@ import { Icon } from '@astrojs/starlight/components';
 interface Props {
 	name: string;
 	twitter?: string;
-    linkedin?: string;
-    github?: string;
+  linkedin?: string;
+  github?: string;
+  website?: string;
```
ajitzero, line 9: Looks like Prettier didn't run here

tomalaforge (approved):

Issue #196, opened 2023-12-07
[Suggestion] Make signalSlice sources lazy load


keithotto:
I noticed in my service where I had created 4 signalSlice objects that they all eagerly fetch their data from their source immediately when the service is first created.

These slices are used in different pages of the website, and it's entirely possible one or more of them may never be called on if the user never visits those pages, so it seems wasteful to be loading that data in advance. 

Would it be possible to delay subscribing to the source observable until the first time a slice has been accessed?

nartc:
Thanks for the suggestion @keithotto 

It'd be tricky to make `signalSlice` sources lazy-load; there might be cases where a configurable `eager` or `lazy` is desired. So, we probably need some time to think about the API for this one.

To work around this, you can start your sources with a subject/many subjects and then the subject when you want to start the sources.

```ts
// instead of
signalSlice({
  sources: [mySource$]
});

// maybe?
const trigger = new Subject();
signalSlice({
  sources: [trigger.pipe(switchMap(() => mySource$))]
})
```

I know it's verbose, but I can't think of anything else off the top of my head.

joshuamorony:
If anyone has any comments I have a little prototype here: https://github.com/joshuamorony/ngxtension-platform/blob/feature/signal-slice-lazy-sources/libs/ngxtension/signal-slice/src/signal-slice.ts

Still need to do a bit of refactoring/testing but the general idea is adding `lazySources` to the config and only using `connect` for those sources  when the state object is accessed via using a `Proxy`.

I'll finish this up soon unless anyone has other ideas

joshuamorony:
Opened a draft PR for this - all seems to work fine, but still facing one blocking issue (detailed in the PR)

joshuamorony:
@keithotto v1.6.0 is out with this change included - I've left it undocumented for now, would you mind giving it a go to see if it works well for your use case and let me know?

keithotto:
I will test it out tomorrow for sure! So it's just as you mentioned above to use `lazySources `instead of `sources`?

nartc:
@keithotto yes

keithotto:
Ok I gave 1.6.0 a quick test with lazySources. It's working well, but I found one case that doesn't seem to trigger the lazy loading. It looks like if you call on the top level signalSlice in your code / template it doesn't connect the lazy source. Admittedly your docs recommend binding to the selectors anyways but it might be something people do.

If I have this in my `AppService`:
```
public $tags = signalSlice({
[...]

PR #197, opened 2023-12-09
feat(signal-slice): add lazySources

joshuamorony:
EDIT: The note below has been resolved

NOTE: This is a draft PR due to one remaining issue and I'm not sure if it can be solved with this set up. I think the crux of the issue is that I am relying on a proxy to detect when the state object is accessed in order to trigger the `lazySources` to `connect`:

```ts
  return new Proxy(slice, {
    get(target, property, receiver) {
      if (!lazySourcesLoaded) {
        lazySourcesLoaded = true;
        connectSources(state, lazySources, injector);
      }

      return Reflect.get(target, property, receiver);
    },
  });
```

Although it still seems to work (the observable values are still set into the signal), it triggers the `SIGNAL_WRITE_FROM_ILLEGAL_CONTEXT` error:

```
ERROR Error: NG0600: Writing to signals is not allowed in a `computed` or an `effect` by default. Use `allowSignalWrites` in the `CreateEffectOptions` to enable this inside effects.
```

---

This adds support for `lazySources` as discussed here: https://github.com/nartc/ngxtension-platform/issues/196

It allows specifying `sources` and `lazySources` in the config - both are identical but the `lazySources` will not be supplied to the `connect` function until the state object has been accessed.

I have left this undocumented for now, ideally would like to have it used a little first. I think I also need to do a full sweep through and improve the docs as a whole as well.

### DIFF
--- a/libs/ngxtension/connect/src/connect.ts
+++ b/libs/ngxtension/connect/src/connect.ts
@@ -1,7 +1,12 @@
-import { DestroyRef, Injector, type WritableSignal } from '@angular/core';
+import {
+	DestroyRef,
+	Injector,
+	untracked,
+	type WritableSignal,
+} from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { assertInjector } from 'ngxtension/assert-injector';
-import { Subscription, isObservable, type Observable } from 'rxjs';
+import { isObservable, Subscription, type Observable } from 'rxjs';
 
 export type PartialOrValue<TValue> = TValue extends object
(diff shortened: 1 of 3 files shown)


nartc:
I can help with the issue

nartc:
Check this [commit](https://github.com/nartc/ngxtension-platform/pull/197/commits/43dea948612bfb5c4d0e5cbeee8caa1d6b6f6b07) for fix to issue

joshuamorony:
Perfect, this seems to work, thank you! This should be ready to go now

PR #205, opened 2023-12-14
fix(host-binding): allows attributes to be removed with nullish value…

geometricpanda:
…s as per @HostBinding

### DIFF
--- a/libs/ngxtension/host-binding/src/host-binding.ts
+++ b/libs/ngxtension/host-binding/src/host-binding.ts
@@ -60,12 +60,10 @@ export function hostBinding<T, S extends Signal<T> | WritableSignal<T>>(
 					);
 					break;
 				case 'attr':
-					const v = value ?? null;
-
-					if (v === null) {
+					if (value == null) {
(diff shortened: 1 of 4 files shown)


libs/ngxtension/host-binding/src/host-binding.ts
```diff
@@ -60,7 +60,13 @@ export function hostBinding<T, S extends Signal<T> | WritableSignal<T>>(
 					);
 					break;
 				case 'attr':
-					renderer.setAttribute(element, property, String(value));
+					const v = value ?? null;
+
+					if (v === null) {
```
nartc, line 65: nit: maybe we can just `== null` check for `value` to remove attribute if `value` is `undefined` or `null`?

geometricpanda: I'm not against that

nartc (approved):

nartc:
@all-contributors please add @geometricpanda for code

nartc:
Thank you for the contribution

PR #210, opened 2023-12-27
docs: third arg is the options-obj for computedFrom

Pascalmh:
The `options`-object is currently inside the `pipe`. This moves it to be the 3rd argument of the `computedFrom`-function.


nartc (approved):
🙏

PR #215, opened 2024-01-05
chore: reformat rxEffect example to remove need for horizontal scrolling

Pascalmh:
At the moment there's some horizontal scrolling going on the documentation page.

This should remove it and make the example easier to read (as you don't need to scroll to the right anymore)

image


Pascalmh:
@nartc `nx format:check` fails and `nx format:write` reverts my changes.

Looks like this would need config changes to work (probably prettier?).

docs/src/content/docs/utilities/Operators/rx-effect.md
```diff
@@ -45,10 +45,13 @@ export class Form {
...
 	});
 
-	readonly #toggleLastNameAccess = rxEffect(this.user.controls.firstName.valueChanges, (firstName) => {
-		if (firstName) this.user.controls.lastName.enable();
-		else this.user.controls.lastName.disable();
-	});
+	readonly #toggleLastNameAccess = rxEffect(
+    	this.user.controls.firstName.valueChanges,
+    	(firstName) => {
+			if (firstName) this.user.controls.lastName.enable();
```
eneajaho, line 51:
```suggestion
		  	if (firstName) this.user.controls.lastName.enable();
		  	else this.user.controls.lastName.disable();
```

nartc (approved):

.prettierrc
```diff
@@ -2,5 +2,13 @@
...
-	"plugins": ["prettier-plugin-organize-imports"]
+	"plugins": ["prettier-plugin-organize-imports"],
+	"overrides": [
+		{
+			"files": "*.md",
+			"options": {
+				"printWidth": 80
+			}
+		}
+	]
```
Pascalmh, line 13: ❤️

PR #216, opened 2024-01-05
feat: add createSignal and createComputed helper functions

eneajaho:


libs/ngxtension/create-signal/src/create-signal.ts
```diff
@@ -0,0 +1,64 @@
...
+			get() {
+				return sig();
+			},
+			set(value: T) {
+				sig.set(value);
+			},
+		},
+	});
+
+	return sig as WritableSignal<T> & { value: T };
```
nartc, line 47: nit/suggestion: let's change this to `defineProperty` instead

```suggestion
	return Object.defineProperty(sig, 'value', {
    get: sig.asReadonly(),
    set: sig.set.bind(sig),
	}) as WritableSignal<T> & { value: T };
```

nartc (changes requested):

libs/ngxtension/create-signal/src/create-signal.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	Object.defineProperties(sig, {
+		value: {
+			get() {
+				return sig();
+			},
+		},
+	});
+
+	return sig as Signal<T> & { value: T };
```
nartc, line 55: nit: c'mon boi! why not change this here as well xD

nartc (changes requested):

libs/ngxtension/create-signal/src/create-signal.ts
eneajaho: There's no readonly for computed?

nartc: yep but `defineProperty` and also you can pass a reference to `sig` to `get` instead:

```ts
return Object.defineProperty(sig, 'value', { get: sig })
```

nartc (approved):

PR #219, opened 2024-01-09
feat(NgxSvgSprite): add a directive for rendering symbols of svg sprites

RobbyRabbitman:
A directive for rendering _symbols_ of svg sprites. It is done with the [`use`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use) element.

## Import

```typescript
import { NgxSvgSpriteFragment } from 'ngxtension/svg-sprite';
```

## Usage

In this example the symbol `github` of the [fontawesome](https://fontawesome.com/) svg sprite `fa-brands` is rendered. A symbol is identified by a `fragment`. Learn more about [URLs](https://svgwg.org/svg2-draft/linking.html#URLReference).

```html
<svg fragment="github" sprite="fa-brands"></svg>
```

Without `NgxSvgSpriteFragment`:

```html
<svg viewBox="0 0 496 512">
	<use href="assets/fontawesome/sprites/brands.svg#github"></use>
</svg>
```

### With Directive Composition Api

In your project you can utilize the [Directive Composition Api](https://angular.io/guide/directive-composition-api) to create specific svg sprites.

In this example a _fontawesome brands_ svg sprite is created.

```html
<svg faBrand="github"></svg>
```

```ts
@Directive({
	selector: 'svg[faBrand]',
	standalone: true,
	hostDirectives: [
		{ directive: NgxSvgSpriteFragment, inputs: ['fragment:faBrand'] },
	],
})
export class FaBrandSvg {
	constructor() {
		inject(NgxSvgSpriteFragment).sprite = 'fa-brands';
	}
}
```

## Configuration

In order to render a symbol, sprites have to be provided.

```ts
provideSvgSprites(
	createSvgSprite({
		name: 'fa-brands',
		baseUrl: 'assets/fontawesome/sprites/brands.svg',
	}),
);
```

The `name` property can reference any arbitrary value, but should be unique, since you can register multiple different svg sprites.

The `sprite` input of the `NgxSvgSpriteFragment` should reference the `name` property of a provided sprite.

### Auto View Box

When a symbol of an svg sprite is rendered the `viewBox` attribute or `height` and `width` _should_ be set. The `svg` element does not copy/use the `viewBox` attribute of the symbol in the svg sprite, therefore the svg will have default dimensions of 300x150 px, which is probably not preferred.

Per default when an svg sprite is registered, the svg sprite is fetched with js in addition. `NgxSvgSpriteFragment` will copy the `viewBox` attribute of the symbol to its host.

This behavior can be disabled.

#### Disable via DI

Auto View Box is disabled for the svg sprite.

```ts
provideSvgSprites(
	createSvgSprite({
		name: 'fa-brands',
		baseUrl: 'assets/fontawesome/sprites/brands.svg',
		autoViewBox: false,
	}),
);
```

#### Disable via `autoViewBoxDisabled` Input

Auto View Box is disabled for a `svg` element, when the `autoViewBoxDisabled` input is set to `false`.

```html
<svg fragment="github" sprite="fa-brands" autoViewBoxDisabled></svg>
```

#### Disable via `viewBox` Attribute

Auto View Box is disabled for a `svg` element, when the `viewBox` attribute already is defined.

```html
<svg fragment="github" sprite="fa-brands" viewBox="0 0 32 32"></svg>
```

### Classes

When the `classes` function is set, a list of classes will be added by the `NgxSvgSpriteFragment` to its host.

```ts
provideSvgSprites(
	createSvgSprite({
		name: 'my-sprite',
		baseUrl: 'path/to/my/sprite.svg',
		classes: (fragment) => ['some-class', `some-other-class-${fragment}`],
	}),
);
```

### Url

Per default when using the `createSvgSprite` function, the `url` will return `'${baseUrl}#${fragment}'`. This can be overwritten:

```ts
provideSvgSprites(
	createSvgSprite({
		name: 'my-sprite',
		baseUrl: 'path/to/my/sprite.svg',
		url: (baseUrl, fragment) => `${baseUrl}#some-prefix-${fragment}`,
	}),
);
```


RobbyRabbitman:
@nartc Hello again, could you please review this feature and share your thoughts on the API 😄

RobbyRabbitman:
@ajitzero The configuration accepts any url. That means any svg can be used. The example shows fontawesome. I think the js fetch is already lazy, I should add test to verify that. The directive adds an use element, I don't know if the use element has a lazy loading feature like the img element. I will research that.

libs/ngxtension/svg-sprite/src/svg-sprite.ts
```diff
@@ -0,0 +1,471 @@
...
+	 */
+	public readonly register = (sprite: NgxSvgSprite) => {
+		this.sprites[sprite.name] = {
+			...sprite,
+			svg$: from(
+				this.ngZone.runOutsideAngular(() => fetch(sprite.baseUrl)),
+			).pipe(
+				switchMap((response) => response.text()),
+				map((text) => {
+					const svg = new DOMParser()
```
nartc, line 111: nit/question: I'm not too familiar with `DOMParser`. Can we just instantiate `DOMParser` once instead?

libs/ngxtension/svg-sprite/src/svg-sprite.ts
```diff
@@ -0,0 +1,471 @@
...
+	/**
+	 * @ignore
+	 */
+	private readonly sprites = inject(NgxSvgSprites);
+
+	/**
+	 * @ignore
+	 */
+	public ngOnInit() {
+		runInInjectionContext(this.injector, () => {
```
nartc, line 310: comment (non-blocking): there's `autoEffect` in `ngxtension/auto-effect` that you can use if you don't want to inject the `Injector` and `runInInjectionContext` manually 😛

libs/ngxtension/svg-sprite/src/svg-sprite.ts
```diff
@@ -0,0 +1,471 @@
...
+
+				// Clear child nodes and remove old classes of this svg.
+				beforeEach(() => {
+					element.replaceChildren();
+					element.classList.remove(...classes);
+				});
+
+				if (fragment == null || sprite == null || spriteConfig == null) return;
+
+				const useElement = document.createElementNS(
```
nartc, line 352: nit: let's `inject(DOCUMENT)` for this

libs/ngxtension/svg-sprite/src/svg-sprite.ts
```diff
@@ -0,0 +1,471 @@
...
+	 * @see {@link NgxSvgSprite.name}
+	 */
+	public readonly sprite$ = signal<string | undefined>(undefined);
+
+	/**
+	 * Whether `autoViewBox` is disabled.
+	 *
+	 * @see overrides {@link NgxSvgSprite.autoViewBox}
+	 */
+	public readonly autoViewBoxDisabled$ = signal<boolean>(false);
```
nartc, line 400: nit: explicit type is redundant

libs/ngxtension/svg-sprite/src/svg-sprite.ts
```diff
@@ -0,0 +1,471 @@
...
+
+	/**
+	 * @ignore
+	 */
+	private readonly spriteConfig$ = computed(() => {
+		const sprite = this.sprite$();
+
+		if (sprite != null) return this.sprites.get(sprite);
+
+		return undefined;
```
nartc, line 410: suggestion: since `this.sprites.get()` returns the sprite from a Record, we won't get a runtime error and will just get `undefined` if the sprite isn't there. Feel free to skip this suggestion though if you think explicitly checking for null is better.

```suggestion
		return this.sprites.get(this.sprite$());
```

libs/ngxtension/svg-sprite/src/svg-sprite.ts
```diff
@@ -0,0 +1,471 @@
...
+
+	/**
+	 * @ignore
+	 */
+	private readonly svg$ = toSignal(
+		toObservable(this.spriteConfig$).pipe(
+			filterNil(),
+			switchMap(({ svg$ }) => svg$),
+		),
+	);
```
nartc, line 421: comment (non-blocking): I almost believe we can use `computedFrom` for this?

libs/ngxtension/svg-sprite/src/svg-sprite.ts
```diff
@@ -0,0 +1,471 @@
...
+	 */
+	public readonly get = (name: string) => this.sprites[name];
+}
+
+/**
+ *
+ * @param sprites
+ * @returns an environment provider which registers icon sprites.
+ */
+export const provideSvgSprites = (...sprites: NgxSvgSprite[]) =>
```
nartc, line 140: nit/suggestion: I think `provideSvgSprites([createSvgSprite()])` is a little verbose. Can't we accept literal objects from the consumers and run `createSvgSprite` internally? In other words, keep `createSvgSprite` an internal API (so one less thing the consumers should care about)

nartc (changes requested):
Thank you for the contributions. Couple of nits and suggestions.

libs/ngxtension/svg-sprite/src/svg-sprite.ts
RobbyRabbitman: I refactored this part from fetch to xhr, because I found out xhr can parse documents. DOMParser is not needed with xhr.

libs/ngxtension/svg-sprite/src/svg-sprite.ts
RobbyRabbitman: Awesome!

libs/ngxtension/svg-sprite/src/svg-sprite.ts
RobbyRabbitman: I like the check better, because `null` and `undefined` are coerced to strings. A sprite _can_ be registered with the name `"null"`or `"undefined"`.

libs/ngxtension/svg-sprite/src/svg-sprite.ts
RobbyRabbitman: Yes, its a use case! I refactored this part.

nartc (approved):

RobbyRabbitman:
@ajitzero I think possible solutions for lazy loading are Intersection Observers: You can even use Angulars new [deferrable views](https://angular.io/guide/defer#on-viewport) to skip setting up the observers by yourself. Maybe like so: 

```html
@defer (on viewport) {
    <svg faBrand="github"></svg>
} @placeholder {
    <span>Some skeletton</span>
}
```

Issue #228, opened 2024-01-15
feat: computedAsync


eneajaho:
Computed that automatically unsubscribes from promises and observables

PR #229, opened 2024-01-15
feat: added computed-async impl

eneajaho:
Closes https://github.com/nartc/ngxtension-platform/issues/228 

Can be used as: 

```ts
  private http = inject(HttpClient);

  productId = injectQueryParams('productId');

  product = computedAsync(() => {
    const productId = this.productId();
    if (!productId) return null;

    return this.http.get<{ id: string; title: string }>(`${API}/${productId}`)

  });
```

> Uses `switchMap` to cancel ongoing requests. 

Playground: [https://stackblitz.com/edit/stackblitz-starters-pzxg8a?file=src%2Fcomputed-async.ts,src%2Fmain.ts,src%2Fglobal_styles.css](https://stackblitz.com/edit/stackblitz-starters-pzxg8a?file=src%2Fcomputed-async.ts,src%2Fmain.ts,src%2Fglobal_styles.css)

### DIFF
--- a/libs/ngxtension/computed-async/src/computed-async.spec.ts
+++ b/libs/ngxtension/computed-async/src/computed-async.spec.ts
@@ -112,17 +112,20 @@ describe(computedAsync.name, () => {
 				expect(logs).toEqual([1, 2]);
 			});
 		}));
-		it('cancels previous computation', fakeAsync(() => {
+		it('behavior: switch (default) -> previous computation', fakeAsync(() => {
 			TestBed.runInInjectionContext(() => {
 				const logs: number[] = [];
 				const value = signal(1);
 
-				const s = computedAsync(() => {
-					return of(value()).pipe(
-						delay(100),
-						tap((v) => logs.push(v)),
-					);
-				});
+				const s = computedAsync(
+					() => {
+						return of(value()).pipe(
+							delay(100),
+							tap((v) => logs.push(v)),
+						);
+					},
+					// { behavior: 'switch' },
+				);
 
 				expect(s()).toEqual(null); // initial value
 				TestBed.flushEffects();
@@ -152,6 +155,139 @@ describe(computedAsync.name, () => {
 				expect(logs).toEqual([1, 3]);
 
 				// 2 was skipped -> the computation was cancelled
+			});
+		}));
+		it('behavior: concat -> does not cancel previous computation', fakeAsync(() => {
+			TestBed.runInInjectionContext(() => {
+				const logs: number[] = [];
+				const value = signal(1);
+
+				const s = computedAsync(
+					() => {
+						return of(value()).pipe(
+							delay(100),
+							tap((v) => logs.push(v)),
+						);
+					},
+					{ behavior: 'concat' },
+				);
+
+				expect(s()).toEqual(null); // initial value
+				TestBed.flushEffects();
+				expect(s()).toEqual(null); // initial value
+				tick(50);
+				expect(s()).toEqual(null); // initial value
+				tick(50);
+				// now we have 100ms passed
+				expect(s()).toEqual(1);
+				expect(logs).toEqual([1]);
+
+				value.set(2);
+				TestBed.flushEffects();
+				expect(s()).toEqual(1); // still the old value
+				tick(50); // wait 50ms
+				expect(s()).toEqual(1);
+				expect(logs).toEqual([1]);
+
+				value.set(3);
+				TestBed.flushEffects();
+
+				tick(50); // wait 50ms
+				expect(s()).toEqual(2);
+				expect(logs).toEqual([1, 2]);
+
+				// now the next computation starts again, so we need to wait 100ms
+				tick(100); // wait 50ms
+				expect(s()).toEqual(3);
+				expect(logs).toEqual([1, 2, 3]);
+			});
+		}));
+		it('behavior: merge -> runs everything in parallel', fakeAsync(() => {
+			TestBed.runInInjectionContext(() => {
+				const logs: number[] = [];
+				const value = signal(1);
+
+				const s = computedAsync(
+					() => {
+						return of(value()).pipe(
+							delay(100),
+							tap((v) => logs.push(v)),
+						);
+					},
+					{ behavior: 'merge' },
+				);
+
+				expect(s()).toEqual(null); // initial value
+				TestBed.flushEffects();
+				expect(s()).toEqual(null); // initial value
+				tick(50);
+				expect(s()).toEqual(null); // initial value
+				tick(50);
+				// now we have 100ms passed
+				expect(s()).toEqual(1);
+				expect(logs).toEqual([1]);
+
+				value.set(2);
+				TestBed.flushEffects();
+
+				value.set(0);
+				TestBed.flushEffects();
+
+				value.set(3);
+				TestBed.flushEffects();
+
+				value.set(4);
+				TestBed.flushEffects();
+
+				tick(100); // wait 50ms
+				expect(s()).toEqual(4);
+				expect(logs).toEqual([1, 2, 0, 3, 4]);
+			});
+		}));
+		it('behavior: exhaust -> skips new computations until last one completes', fakeAsync(() => {
+			TestBed.runInInjectionContext(() => {
+				const logs: number[] = [];
+				const value = signal(1);
+
+				const s = computedAsync(
+					() => {
+						return of(value()).pipe(
+							delay(100),
+							tap((v) => logs.push(v)),
+						);
+					},
+					{ behavior: 'exhaust' },
+				);
+
+				expect(s()).toEqual(null); // initial value
+				TestBed.flushEffects();
+				expect(s()).toEqual(null); // initial value
+				tick(50);
+				expect(s()).toEqual(null); // initial value
+				tick(50);
+				// now we have 100ms passed
+				expect(s()).toEqual(1);
+				expect(logs).toEqual([1]);
+
+				value.set(2);
+				TestBed.flushEffects();
+				expect(s()).toEqual(1); // still the old value
+				tick(50); // wait 50ms
+
+				value.set(3);
+				TestBed.flushEffects();
+
+				expect(s()).toEqual(1);
+				expect(logs).toEqual([1]);
+
+				tick(50); // wait 50ms
+				expect(s()).toEqual(2);
+				expect(logs).toEqual([1, 2]);
+
+				// now the next computation starts again, so we need to wait 100ms
+				tick(100); // wait 50ms
+				expect(s()).toEqual(2);
+				expect(logs).toEqual([1, 2]);
 			});
 		}));
 	});
--- a/libs/ngxtension/computed-async/src/computed-async.spec.ts
+++ b/libs/ngxtension/computed-async/src/computed-async.spec.ts
@@ -158,34 +158,36 @@ describe(computedAsync.name, () => {
 
 	describe('is typesafe', () => {
 		it('initial value', () => {
-			const a: Signal<number> = computedAsync(() => {
-				if (Math.random() > 0.5) return Promise.resolve(1);
-				return Promise.resolve(1);
+			TestBed.runInInjectionContext(() => {
+				const a: Signal<number> = computedAsync(() => {
+					if (Math.random() > 0.5) return Promise.resolve(1);
+					return Promise.resolve(1);
+				});
+
+				const b: Signal<number> = computedAsync(() => {
+					if (Math.random() > 0.5) return of(1);
+					return of(1);
+				});
+
+				const c: Signal<number | null> = computedAsync(
+					() => {
+						return 1;
+					},
+					{ initialValue: null },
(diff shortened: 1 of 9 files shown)


ilirbeqirii:
Can we add the _**null**_ value as default if the initialValue is not provided just like here but also for **computedFrom** function?

eneajaho:
> Can we add the _**null**_ value as default if the initialValue is not provided just like here but also for **computedFrom** function?

Probably will make this work the same as computedFrom when it comes to the initial value.

samuelfernandez:
This is great! I was thinking on something similar over the last days, happy to see it in this fantastic repo. A few ideas that could improve the API:

- Async code benefits from knowing the status of the async process. Instead of returning the final result, `computedAsync` would ideally return an object like `{ value: T, status }`, being the `status` something similar to `pending`, `completed` or `failed`. This would enable implementers to show a loading indicator or similar.
- There could be some cases where `switchMap` was not the preferred option, and instead another operator like `exhaustMap` would fit. The options param could have a new optional property `operator` for this purpose, defaulting to `switchMap`
- Even though observables can be cancelled, promises as they are can't. Ideally the function of the `computedAsync` would pass an `AbortController`, similar to how `fetch` works. Example: `computedAsync(({ signal }) => fetch(url, { signal }))`

joshuamorony:
On the topic of optionally specifying the operator, I think a config that does not use the RxJS terminology might be beneficial (we are abstracting away the stream/observable aspect after all)

Perhaps something like:

```ts
  product = computedAsync(() => {
    const productId = this.productId();
    if (!productId) return null;

    return this.http.get<{ id: string; title: string }>(`${API}/${productId}`)

  }, { handleMultiple: 'parallel'});
```

I don't really like `handleMultiple` but something that conveys `how should we handle multiple simultaneous requests`. For the options I'm thinking:

```
'switch' // switchMap - we should handle a new value by switching to it
'race' // mergeMap - we should handle a new value by racing it with the existing one
'queue' // concatMap - we should handle a new value by queueing it after the existing one
'ignore' // exhaustMap - we should ignore any new values if there is an existing one
```

joshuamorony:
> @joshuamorony I don't think we should come up with anything different for Angular ecosystem. We can have an option called `flatten` and folks should be able to pass the operator reference in:
> 
> ```js
> { flatten: switchMap } // default
> { flatten: concatMap }
> { flatten: mergeMap }
> { flatten: exhaustMap }
> ```

I don't have particularly strong opinions on this, but my thinking is that:

1) The path from Angular seems to be to make RxJS completely optional eventually. If we have the user pass in the operator they would still need to be importing from RxJS

2) Same point as above basically, but we have introduced the concept of RxJS rather than abstracting it away completely

3) The naming of the operators are based on the observable stream model, again if we are attempting to provide a signal based abstraction then perhaps we can create more intuitive names not rooted in the mental model of streams

LcsGa:
About the `flatten` suggestion, I thought it could help you to know that the RxJS V8 might implement a knew [`flatMap`](https://github.com/ReactiveX/rxjs/discussions/7429) operator!

When it is released, the bundle size issue shouldn't be a problem anymore if we choose names instead of the operators (as `flatMap` would then be the only one imported internally).

That said I would go for `flatten` being a union type of our choice (like @joshuamorony's first idea) even if, for a little while, it would mean a larger bundle size.

hoc081098:
It could be FlattenStrategy.
I have used it in my Flutter library with rxdart 😂

https://github.com/hoc081098/stream_loader/blob/master/lib%2Fsrc%2Floader_bloc.dart#L15-L27

```dart
enum FlattenStrategy {
  /// uses [FlatMapExtension.flatMap].
  merge,

  /// uses [Stream.asyncExpand].
  concat,

  /// uses [SwitchMapExtension.switchMap].
  latest,

  /// uses [ExhaustMapExtension.exhaustMap].
  first,
}
```

renatoaraujoc:
> @joshuamorony I don't think we should come up with anything different for Angular ecosystem. We can have an option called `flatten` and folks should be able to pass the operator reference in:
> 
> ```js
> { flatten: switchMap } // default
> { flatten: concatMap }
> { flatten: mergeMap }
> { flatten: exhaustMap }
> ```

I like this idea more but instead of passing the imported operator itself, pass a union of operator string names: `'switchMap' | 'concatMap'` etc.

malek-itani:
I would go with having a similar naming convention as RxJS it is better and helps the community to use the extensions library fast 

`behavior?: 'merge' | 'switch' | 'exhaust';`

OmerGronich:
> I would go with having a similar naming convention as RxJS it is better and helps the community to use the extensions library fast
> 
> `behavior?: 'merge' | 'switch' | 'exhaust';`

I really like this one. It would make the usage really simple and clean. And also mirrors the upcoming flatMap operator.

joshuamorony:
> > I would go with having a similar naming convention as RxJS it is better and helps the community to use the extensions library fast
> > `behavior?: 'merge' | 'switch' | 'exhaust';`
> 
> I really like this one. It would make the usage really simple and clean. And also mirrors the upcoming flatMap operator.

The flatMap naming looks like it is going to be intuitive enough anyway, so I agree mirroring that makes sense - it will still be intuitive enough for people not familiar with RxJS.

I think `switch`, `merge`, `ignore`, `queue` would be close to ideal, but if we can sync up the naming with RxJS with `switch`, `merge`, `exhaust`, `concat` then I think that is easily worth it.

eneajaho:
Thanks @ilirbeqirii @samuelfernandez @joshuamorony @nartc @LcsGa @hoc081098 @renatoaraujoc @malek-itani @SerkanSipahi @OmerGronich for the comments 🙌

I will be going with the string approach and follow the same names as rxjs, so it's easier for the devs who already know rxjs. And with some good docs we can make it easy for the new devs using it too.

eneajaho:
I'm also thinking on adding the previousValue in the callback, and this way we can compute state based on previous state (if exists)

eneajaho:
> Nice work! A couple of questions:
> 
> 1. Why not name it `computedToAsync` or even `toAsync` (maybe a bit too generic)?
> 2. For the flattening, it ties you again to RxJS underneath. Are there plans to make non-Observable-based versions of these flattening strategies? If the goal is for RxJS to be optional at some point.

Hi @brandonroberts 

From what I saw from some links posted, Vue community also has the same function (https://vueuse.org/core/computedAsync/). So the name makes sense to me as it is. 


Regarding being tied to rxjs, yes that's on purpose for the moment (because of the whole built-in functionalities that we don't have to re-create). 

For the moment there are no plans to make non-Observable-based versions of them as these 4 operators, most of the time will be already part of the code because of Angular router source code (it includes concatMap, mergeMap, switchMap), so it's not adding that much bundle size.

libs/ngxtension/computed-async/src/computed-async.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+	});
+
+	describe('works with previousValue', () => {
+		it('and can be retrieved in the callback fn', fakeAsync(() => {
+			TestBed.runInInjectionContext(() => {
+				const logs: number[] = [];
+				const value = signal(1);
+
+				// TODO(enea): the prev param has unknown type for some reason
+				const s = computedAsync((previousValue: number | undefined) => {
```
eneajaho, line 68: @nartc the typing of this function is killing me 🤯

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -0,0 +1,158 @@
...
+		const effectRef = effect(
+			() => {
+				// we need to have an untracked() here because we don't want to register the sourceValue as a dependency
+				// otherwise, we would have an infinite loop
+				const currentValue = untracked(() => sourceValue());
+
+				const newSource = computation(currentValue);
+				if (!isObservable(newSource) && !isPromise(newSource)) {
+					// if the new source is not an observable or a promise, we set the value immediately
+					untracked(() => sourceValue.set(newSource));
```
ribizli, line 99: why this `untracked` is needed. I think no signal is called here at all. The same for line 104.

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -0,0 +1,158 @@
...
+		case 'merge':
+			return source.pipe(mergeMap((s) => s));
+		case 'concat':
+			return source.pipe(concatMap((s) => s));
+		case 'exhaust':
+			return source.pipe(exhaustMap((s) => s));
+		default: // switch
+			return source.pipe(switchMap((s) => s));
+	}
+}
```
hoc081098, line 154: I think `mergeAll`, `switchAll`, `exhaustAll` and `concatAll` can be used instead of `mergeMap((s) => s`, ...

libs/ngxtension/computed-async/src/computed-async.ts
eneajaho: You cannot set a signal inside an effect, but looks like untracked helps with that thing too 😄

https://github.com/angular/angular/issues/53986

libs/ngxtension/computed-async/src/computed-async.ts
eneajaho: Hi, thanks.
But, there will be only one observable at a time? How will these other operators change anything?

LcsGa: Hi! It doesn't mean that you flatten multiple observables but that it will flattne every observables going through that stream.

You wrote (let's take swtichMap as an example):
```ts
source.pipe(switchMap((s) => s));
```

But `(s) => s` can be "shortened" with [`identity`](https://rxjs.dev/api/index/function/identity)
```ts
source.pipe(switchMap(identity)); // which is the same as source.pipe(switchMap((s) => s))
```
And `swithMap(identity)` is the [`swithchAll`](https://github.com/ReactiveX/rxjs/blob/7f36e90626ba7f93687e227f8b2bf3c0302ac3ee/packages/rxjs/src/internal/operators/switchAll.ts#L62) operator
```ts
source.pipe(switchAll()); // which is the same as source.pipe(switchMap(identity))
```

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -0,0 +1,62 @@
...
+
+		const sourceResult = source$
+			.pipe(switchMap((source$) => source$))
+			.subscribe({
+				next: (value) => {
+					sourceValue.set(value);
+				},
+				error: (error) => {
+					sourceValue.set(null);
+					throw error;
```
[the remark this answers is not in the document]
LcsGa: Well, I just discovered that I had to `submit` my review for you to see my comments... 😅
They might not be usefull anymore, I'll check that asasp and remove them if they are not, sorry

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -0,0 +1,158 @@
...
+			sourceEvent$,
+			options?.behavior ?? 'switch',
+		);
+
+		const sourceResult = source$.subscribe({
+			next: (value) => sourceValue.set(value),
+			error: (error) => {
+				// NOTE: Error should be handled by the user (using catchError or .catch())
+				sourceValue.set(error);
+				throw error;
```
LcsGa, line 119: There is no need to throw the error here, it is just an error notification. The error is not caught.

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -0,0 +1,158 @@
...
+
+				const newSource = computation(currentValue);
+				if (!isObservable(newSource) && !isPromise(newSource)) {
+					// if the new source is not an observable or a promise, we set the value immediately
+					untracked(() => sourceValue.set(newSource));
+					return;
+				}
+
+				// we untrack the source$.next() so that we don't register other signals as dependencies
+				untracked(() => sourceEvent$.next(newSource));
```
LcsGa, line 104:
```suggestion
				if (!isObservable(newSource) && !isPromise(newSource)) {
					// if the new source is not an observable or a promise, we set the value immediately
					untracked(() => sourceValue.set(newSource));
				} else {
				    // we untrack the source$.next() so that we don't register other signals as dependencies
				    untracked(() => sourceEvent$.next(newSource));
				}
```

libs/ngxtension/computed-async/src/computed-async.ts
eneajaho: Nice! I learned something new and just updated the PR. Thanks 🤩

tomalaforge: Oh nice @LcsGa I didn't know it either. Thanks 🙏

libs/ngxtension/computed-async/src/computed-async.ts
eneajaho: Updated! Thanks 🙏

libs/ngxtension/computed-async/src/computed-async.ts
LcsGa: You're welcome 😁

PR #232, opened 2024-01-20
injectQueryParam refactor

Dafnik:
The currrent `injectQueryParams()`method lacks support for Array query parameter. 
For example: `test.url/search?productId=2&productId=3`.

I'm proposing an API familiar to the new Signal Inputs, still allowing great customization.

```typescript
queryParams = queryParams(); // types to Params
idParam = queryParams('id', { transform: numberAttribute }); // infers type to number | null
idParams = queryParams.multiple('id', { transform: numberAttribute }); // infers type to number[]
searchParams = queryParams.multiple('query'); // types to string[]
paramKeysList = queryParams((params) => Object.keys(params)); // infers type
```

The naming can still be discussed of course.

Would love to hear your thoughts on this :)


ajitzero:
Two minor suggestions/questions:
1. Instead of `number[]`, I feel it should return `number[] | null`. This `null` would be useful for knowing when no values are passed, and we would then never get an empty array. If the result is `number[]` instead, we would still need to check whether the array has at least one element before proceeding.
2. Extremely personal opinion: I find `queryParams.multiple` (your suggestion) & `input.required` (official Angular syntax) a bit confusing (readability), mainly due to lack of familiarity. My approach would be to pass `{ multiple: true }` along with the transform, like so:

```typescript
  #idParams = queryParams('id', { multiple: true });
  // OR
  #idParams = queryParams('id', { multiple: true, transform: numberAttribute });
```
Suggested types: (These are probably wrong or can be simplified further)
```typescript
type BaseQueryParamsOptions<T> = {
  transform?: (input: string): input is T => T;
}
type QueryParamsOptions<T> =
  | ((BaseQueryParamsOptions<T> & { multiple: false; }) => T | null)
  | ((BaseQueryParamsOptions<T> & { multiple: true; }) => T[] | null);
```

libs/ngxtension/inject-query-params/src/inject-query-params.spec.ts
```diff
@@ -1,12 +1,25 @@
-import { Component } from '@angular/core';
+import { Component, numberAttribute } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 
 import { provideRouter } from '@angular/router';
 import { RouterTestingHarness } from '@angular/router/testing';
-import { injectQueryParams } from './inject-query-params';
+import { queryParams } from './inject-query-params';
```
ajitzero, line 6: Minor: I think following the `inject` prefix for CIFs is more intentional/consistent.

libs/ngxtension/inject-query-params/src/inject-query-params.spec.ts
```diff
@@ -22,22 +35,96 @@ describe(injectQueryParams.name, () => {
...
 });
 
-@Component({
-	standalone: true,
-	template: ``,
-})
-export class SearchComponent {
-	queryParams = injectQueryParams();
-	searchParam = injectQueryParams('query');
-	paramKeysList = injectQueryParams((params) => Object.keys(params));
```
ajitzero, line 42: Question: Is `queryParams.multiple` a simpler variant of `injectQueryParams((params) => Object.keys(params))`?

Dafnik: No, its more like if you know you want all query params of a specific key as an array!

For example you've got the url `/search?productId=1&productId=2`.
Now you want the productIds as an array. With the current implementation `#idParam = injectQueryParams('productId')` you get the type `string|null` even when it should be `string[]|null`. You have no way to tell the method you want an array.

Perhaps `injectQueryParams.array()` would be a better name?

libs/ngxtension/inject-query-params/src/inject-query-params.spec.ts
Dafnik: Works for me. I just thought it was really simple and resembled the angular input signals :)

Dafnik:
Thanks for the feedback! :)

> Instead of number[], I feel it should return number[] | null. This null would be useful for knowing when no values are passed, and we would then never get an empty array. If the result is number[] instead, we would still need to check whether the array has at least one element before proceeding.

Yeah, I also thought about this. I was generally more against it (as you need to always use the `?`operator now to access it.
But with your point in mind it makes more sense to make it nullable.

One thing I wasn't sure about is, why is it using `null` and not `undefined`? is there a particular reason for it as I find `undefined` most of the time better

> Extremely personal opinion: I find queryParams.multiple (your suggestion) & input.required (official Angular syntax) a bit confusing (readability), mainly due to lack of familiarity. My approach would be to pass { multiple: true } along with the transform, like so

I wasn't sure about this either 😅 It lacks of familiarity for sure but it makes the typing much easier (I think). I would perhaps also go with `isArray`, or `array` if we want to go with it being an option.


Generally speaking, I also thought about adding the possibility of an initialValue like input signals.

ajitzero:
> One thing I wasn't sure about is, why is it using `null` and not `undefined`? is there a particular reason for it as I find `undefined` most of the time better

Oh sure, I personally prefer just one of those, either `null` or `undefined`. There's [this discussion](https://github.com/sindresorhus/meta/discussions/7) which supports having just `undefined`.

Switching `null` to `undefined` is fine, but it would technically be a minor breaking change. It would be good to get @nartc's stance on this.

> but it makes the typing much easier (I think)

Probably! This one's more about unfamiliarity, but with Angular officially using this format, it's a non-issue. Having all configurations in a single object seems slightly more convenient but is probably more in the personal preference category. We can retain your original change and ignore my comment on that part.

libs/ngxtension/inject-query-params/src/inject-query-params.spec.ts
eneajaho: There is a small diff between inputs and this fn. Inputs don't inject anything under the hood, while this function does. And if the inject prefix is not used, devs may think that it's safe to use directly in a non injection context, but that would throw an error. 

So the naming is intentional here.

Dafnik:
Do you know why Angular the required value from object to this extra function? Perhaps it was used that often, that it legitimates its own function.

I'm not really against objects for configs generally speaking :)

nartc:
@Dafnik Is this a draft or is it ready?

Dafnik:
No its still a draft. Docs are still missing, some code needs to be changed and I'm not sure about the naming on the array method either.

Like if it should be `injectQueryParams.array()` or `.multiple`

libs/ngxtension/inject-query-params/src/inject-query-params.ts
```diff
@@ -3,55 +3,269 @@ import { toSignal } from '@angular/core/rxjs-interop';
...
- * @param transform
+ * The `injectQueryParams` function allows you to access and manipulate query parameters from the current route.
+ *
+ * @param {string} key - The name of the query parameter to retrieve.
+ * @param {QueryParamsOptions} options - Optional configuration options for the query parameter.
+ * @returns {Signal} A `Signal` that emits the transformed value of the specified query parameter, or `null` if it's not present.
+ */
+export function injectQueryParams(
+	key?: string,
+	options?: undefined,
```
JeanMeche, line 54: This parameter could be removed right  ?

libs/ngxtension/inject-query-params/src/inject-query-params.ts
```diff
@@ -3,55 +3,269 @@ import { toSignal } from '@angular/core/rxjs-interop';
...
  * const queryParams = injectQueryParams(); // returns the entire query params object
- *
- * @param keyOrTransform OPTIONAL The key of the query param to return, or a transform function to apply to the query params object
  */
-export function injectQueryParams<T>(
-	keyOrTransform?: string | ((params: Params) => T),
-): Signal<T | Params | string | null> {
+export function injectQueryParams<ReadT>(
+	keyOrParamsTransform?: string | ((params: Params) => ReadT),
+	options?: QueryParamsOptions<ReadT, string, ReadT>,
```
JeanMeche, line 125:
```suggestion
	options: QueryParamsOptions<ReadT, string, ReadT> = {},
```
It would remove the condition a few lines below.

libs/ngxtension/inject-query-params/src/inject-query-params.ts
```diff
@@ -3,55 +3,269 @@ import { toSignal } from '@angular/core/rxjs-interop';
...
+	/**
+	 * Retrieve an array query parameter with optional configuration options.
+	 *
+	 * @param {string} key - The name of the array query parameter to retrieve.
+	 * @param {QueryParamsOptions} options - Optional configuration options for the array query parameter.
+	 * @returns {Signal} A `Signal` that emits an array of values for the specified query parameter, or `null` if it's not present.
+	 */
+	export function array(
+		key: string,
+		options?: undefined,
```
JeanMeche, line 183:
```suggestion
```
nit: this could be removed

Dafnik:
Thanks for the review, implemented all suggestions!

Dafnik:
H there! Is there anything missing? @nartc

docs/src/content/docs/utilities/Injectors/inject-query-params.md
```diff
@@ -67,3 +67,47 @@ class TestComponent {
...
+	`,
+})
+class TestComponent {
+	number1 = injectQueryParams('id', { transform: numberAttribute }); // returns a signal with the value of the search query param
+
+	newNumber = computed(() => this.number1() * 2);
+}
+```
+
+If we want to get the values for a specific query param, we can pass the name of the query param, to `injectQueryParams.array`.
```
eneajaho, line 86: I'd move this to a new section in docs.

nartc:
> H there! Is there anything missing? @nartc 

yes, my attention 😂. I'll take a final look today

eneajaho:
@Dafnik I left a comment, and I'd like to merge this. Can you rebase please and also take a look at the changes that were done to injectQueryParams functions, and do the same thing for some places in the PR?

Dafnik:
Hi, what you mean exactly by this?

> take a look at the changes that were done to injectQueryParams functions, and do the same thing for some places in the PR?

eneajaho:
> Hi, what you mean exactly by this?
> 
> > take a look at the changes that were done to injectQueryParams functions, and do the same thing for some places in the PR?

This PR will be merged soon https://github.com/nartc/ngxtension-platform/pull/272 and has some changes to the way the utility works under the hood.

Dafnik:
I still don't quite get it 😅

What do you want me exactly to do with the changes made in your pr? Should I cherry-pick the commit so the changes are also in this pr?
Or should I apply the changes from your pr manually?

Still happy to do it and helo though :)

Dafnik:
Is this still considered for inclusion?

docs/src/content/docs/utilities/Injectors/inject-query-params.md
```diff
@@ -45,25 +63,136 @@ If we want to get the value for a specific query param, we can pass the name of
 class TestComponent {
 	searchParam = injectQueryParams('search'); // returns a signal with the value of the search query param
 
-	filteredUsers = computedFrom(
-		[this.searchParam],
-		switchMap((searchQuery) =>
-			this.userService.getUsers(searchQuery).pipe(startWith([])),
-		),
+	filteredUsers = computedAsync(
+		() => this.userService.getUsers(searchQuery ?? ''),
```
eneajaho, line 67:
```suggestion
		() => this.userService.getUsers(this.searchParam() ?? ''),
```

docs/src/content/docs/utilities/Injectors/inject-query-params.md
```diff
@@ -45,25 +63,136 @@ If we want to get the value for a specific query param, we can pass the name of
...
 ```ts
-@Component()
+@Component({
+	template: `
+		Number: {{ number1() }} New number: {{ newNumber() }}
+	`,
+})
 class TestComponent {
-	queryParamsKeys = injectQueryParams((params) => Object.keys(params)); // returns a signal with the keys of the query params
+	number1 = injectQueryParams('id', { transform: numberAttribute }); // returns a signal with the value of the search query param
```
eneajaho, line 82: I'd like to have examples which are meaningful. So let's convert this to a pageNumber and use better var names.

```suggestion
	pageNumber = injectQueryParams('pageNumber', { transform: numberAttribute });
```

docs/src/content/docs/utilities/Injectors/inject-query-params.md
```diff
@@ -45,25 +63,136 @@ If we want to get the value for a specific query param, we can pass the name of
...
 
-	allQueryParamsArePassed = computed(() => {
-		const keys = this.queryParamsKeys();
-		return ['search', 'sort', 'page'].every((x) => keys.includes(x));
+	newNumber = computed(() => this.number1() * 2);
```
eneajaho, line 84:
```suggestion
```

docs/src/content/docs/utilities/Injectors/inject-query-params.md
```diff
@@ -45,25 +63,136 @@ If we want to get the value for a specific query param, we can pass the name of
...
+		} @empty {
+			<div>No users!</div>
+		}
+	`,
+})
+class TestComponent {
+	// returns a signal with the value of the search query param or '' if not provided.
+	searchParam = injectQueryParams('search', { initialValue: '' });
+
+	filteredUsers = computedAsync(() => this.userService.getUsers(searchQuery), {
```
eneajaho, line 106:
```suggestion
	filteredUsers = computedAsync(() => this.userService.getUsers(this.searchParam()), {
```

eneajaho (changes requested):

eneajaho (approved):

eneajaho:
Thanks @Dafnik 🚀

PR #234, opened 2024-01-21
fix(connect): update literal object values correctly for non-literal object values

joshuamorony:
This fixes an issue where the signal would not be updated correctly when a non-literal object value is used to update a literal object value.

For example, attempting to update `{}` with `null` would result in the value still being `{}`


joshuamorony:
I updated the PR to also handle the situation where a value is being updated from an object to a non-literal object via a reducer

nartc (approved):

PR #237, opened 2024-01-22
fix: add dependency ts-morph for Signals Input migrator

rainerhahnekamp:
Add ts-morph as a dependency, which is needed for the Signals Input migrator.


nartc:
Thanks for the PR. Not sure why I missed this. Even `ng-packagr` was updated to have `ts-morph` x.x 

@all-contributors please add @rainerhahnekamp for code

nartc (approved):

PR #239, opened 2024-01-22
feat(create-injectable): add createInjectable (replaces createService)

joshuamorony:
Discussion: https://twitter.com/spierala/status/1748639069171568771

This PR removes the undocumented `createService` function added to the `create-injection-token` entry point in favour of:

1) Moving it to its own entry point
2) Renaming it to `createInjectable`
3) Having it return injectable directly instead of the injectFn/provideFn/token tuple that `createInjectionToken` returns by default

I've marked this as experimental for now to give us some flexibility in potential changes

EDIT: Updated usage based on updates from @nartc 

## Non-root Service

```ts
// defining a service
export const MyService = createInjectable(() => {
	const myState = signal(1);

	return { myState: myState.asReadonly() };
});
```

```ts
// provide the service
{
	providers: [MyService];
}
```

```ts
// using the service
const myService = inject(MyService);
```

## Root Service

```ts
// defining a root service
export const MyService = createInjectable(
	() => {
		const myState = signal(1);
		return { myState: myState.asReadonly() };
	},
	{ providedIn: 'root' },
);
```

```ts
// using the service
const myService = inject(MyService);
```

@nartc I remember briefly discussing changing the `{ isRoot: false }` option in `createInjectionToken` - if that is still happening we could add a change to `createInjectable` to pre-empt whatever that is going to be?

### DIFF
--- a/docs/src/content/docs/utilities/Injectors/create-injectable.md
+++ b/docs/src/content/docs/utilities/Injectors/create-injectable.md
@@ -19,9 +19,24 @@ configuration.
 
 ## Usage
 
+### Shared Service
+
 ```ts
-// defining a service
-export const MyService = createInjectable(
+// defining a shared service
+export const [MyService] = createInjectable(() => {
+	const myState = signal(1);
+
+	return { myState: myState.asReadonly() };
+});
+
+// using the service
+const myService = inject(MyService);
+```
+
+### Non-root Service
+
+```ts
+export const [MyService, provideMyService] = createInjectable(
 	() => {
 		const myState = signal(1);
 
@@ -29,7 +44,14 @@ export const MyService = createInjectable(
 	},
 	{ isRoot: false },
 );
+```
 
-// using a service
+```ts
+{
+	providers: [provideMyService()];
+}
+```
+
+```ts
 const myService = inject(MyService);
 ```
--- a/docs/src/content/docs/utilities/Injectors/create-injectable.md
+++ b/docs/src/content/docs/utilities/Injectors/create-injectable.md
@@ -6,52 +6,54 @@ badge: experimental
 contributor: josh-morony
 ---
 
-`createInjectable` uses `createInjectionToken` behind the scenes as a way to
-create injectable services and other types of injectables in Angular without
-using classes and decorators.
+`createInjectable` returns a class behind the scenes as a way to
+create injectable services and other types of injectables in Angular without using classes and decorators.
 
 The general difference is that rather than using a class, we use a `function` to
 create the injectable. Whatever the function returns is what will be the
 consumable public API of the service, everything else will be private.
 
-To create an injectable that is `providedIn: 'root'` you can omit the `isRoot`
-configuration.
+Pass `{ providedIn: 'root' }` to the 2nd argument of `createInjectable` if you want to create a root service.
 
 ## Usage
-
-### Shared Service
-
-```ts
-// defining a shared service
-export const [MyService] = createInjectable(() => {
-	const myState = signal(1);
-
-	return { myState: myState.asReadonly() };
-});
-
-// using the service
-const myService = inject(MyService);
-```
 
 ### Non-root Service
 
 ```ts
-export const [MyService, provideMyService] = createInjectable(
-	() => {
-		const myState = signal(1);
+// defining a service
+export const MyService = createInjectable(() => {
+	const myState = signal(1);
 
-		return { myState: myState.asReadonly() };
-	},
-	{ isRoot: false },
-);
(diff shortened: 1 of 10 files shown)


nartc:
re:`isRoot` I want to change it to `providedIn` so it's consistent with `Injectable` decorator

libs/ngxtension/create-injectable/src/create-injectable.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function createInjectable<
+	TFactory extends (...args: any[]) => object,
+	TFactoryDeps extends Parameters<TFactory> = Parameters<TFactory>,
+	TOptions extends CreateInjectionTokenOptions<
+		TFactory,
+		TFactoryDeps
+	> = CreateInjectableOptions<TFactory, TFactoryDeps>,
+>(factory: TFactory, options?: TOptions): InjectionToken<ReturnType<TFactory>> {
+	return createInjectionToken(factory, options)[2];
```
nartc, line 31: this won't work because `[2]` is just an `InjectionToken`. It cannot be used on its own in the `providers` array. We'd need a different implementation or adjust `createInjectionToken` somehow

nartc (changes requested):

libs/ngxtension/create-injectable/src/create-injectable.ts
joshuamorony: I should've added a test for provide, will work on that thanks

joshuamorony: For now I've just changed it to return the provide function as well so that we have something working, but perhaps we want to leave the exact implementation open to discussion still

nartc: Sure thing.

nartc: Pushed my implementation. If you don't like it, we can revert.

joshuamorony: Looks good to me - I'll update the docs shortly. Do we want to leave the PR open and I can invite people from the video to comment/discuss? or we can just leave it as experimental for now

nartc: I've updated docs. I think we can merge it as-is. If feedbacks start coming in, we can always re-iterate

docs/src/content/docs/utilities/Injectors/create-injectable.md
```diff
@@ -0,0 +1,57 @@
...
+To create an injectable that is `providedIn: 'root'` you can omit the `isRoot`
+configuration.
+
+## Usage
+
+### Shared Service
+
+```ts
+// defining a shared service
+export const [MyService] = createInjectable(() => {
```
spierala: Looking at the code, I think that `createInjectable` does not return an Array anymore.

libs/ngxtension/create-injectable/src/create-injectable.ts
nartc: Adjusted tests to showcase the behavior or root service vs non-root service

joshuamorony: Sounds good! Nice work!

nartc:
@joshuamorony if you're ok with the implementation, can you revert the change in `create-injection-token`? And add yourself as contributor please, I wouldn't have jumped in here if you didn't open the PR in the first place. Don't downplay your contribution

libs/ngxtension/create-injectable/src/create-injectable.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+				count += 1;
+				return { someProp: 1 };
+			},
+			{ providedIn: 'root' },
+		);
+
+		TestBed.runInInjectionContext(() => {
+			// should be lazy until `inject()` is invoked
+			expect(count).toEqual(0);
+			const service = inject(MyInjectable);
```
spierala, line 19: Maybe it could be useful to inject `MyInjectable` two times... and then check that count does not increment anymore after the first injection.

Then we are quite sure that it is a Singleton service.

Same thing could be done for the non-root injectable, but there we would expect that the count is incremented for every injection.

nartc:
> > @joshuamorony if you're ok with the implementation, can you revert the change in `create-injection-token`? And add yourself as contributor please, I wouldn't have jumped in here if you didn't open the PR in the first place. Don't downplay your contribution
> 
> 
> 
> Done and done - perhaps I can put up another PR to enable supplying multiple contributors in the docs?

That'd be great

PR #242, opened 2024-01-23
docs: add support for multiple contributors to a utility

joshuamorony:
This adds support for supplying multiple contributors to a specific utility for cases where something is co-created by more than one person.

I have migrated the existing format:

```
contributor: josh-morony
```

to take an array instead:

```
contributors: ['josh-morony']
```

or

```
contributors: ['josh-morony', 'chau-tran']
```

I've also updated the styling so that utilities with multiple creators + with or without social links still looks nice.

As well as migrating existing docs, this PR also:
* Adds `chau-tran` as creator for `createInjectable`
* Updates social links for `josh-morony`


docs/src/components/Contributor.astro
```diff
@@ -1,59 +1,75 @@
...
+<p class="contributors">
+	Created by
+
 	{
-		website && (
-			<a href={website}>
-				<Icon class="icon" name="external" size="0.75rem" />
-			</a>
+		contributors.map(
+			({ data: { name, twitter, linkedin, github, website } }) => (
```
ajitzero, line 24: Minor/question: Do we need to mention `data` in the type?

`contributors` is supposed to have `Contributor[]` type, but seems to have `{ data: Contributor[] }` instead. Is this something specific to Astro?

joshuamorony: Using 'getEntry' to get a contributor from the collection returns the data in this format:

```
  {
    id: 'chau-tran',
    collection: 'contributors',
    data: {
      name: 'Chau Tran',
      twitter: 'https://twitter.com/Nartc1410',
      github: 'https://github.com/nartc'
    }
  }
```

(the component is passed an array of these, which is why I am mapping over it and destructuring on data)

I don't actually use Astro so I'm not sure if there is a nicer way to go about it or not

ajitzero: (typing from phone so can't test this myself) - Not 100% sure, but I think the type _could_ use something like this (copied from one of my repos):

```ts
import type { CollectionEntry } from 'astro:content';
// interface Props {
// 	 page: Page<CollectionEntry<'contributor'>>;
// }
// OR
interface Props {
	contributors: Array<CollectionEntry<'contributor'>>;
}
```

ajitzero (approved):

docs/src/components/Content.astro
```diff
@@ -4,12 +4,16 @@ import type { Props } from '@astrojs/starlight/props';
...
+	? await Promise.all(
+			Astro.props.entry.data.contributors.map((contributor) =>
+				getEntry(contributor),
+			),
+		)
 	: null;
 ---
 
-{contributor && <Contributor {...contributor.data} />}
+{contributors && <Contributor {contributors} />}
```
nartc, line 16: nit: rename the component to `Contributors`

nartc (approved):

PR #247, opened 2024-01-23
fix(computedAsync): typings should return a union of undefined when no initial value.

### LINKED ISSUE #243: computedAsync must return T | undefined, but returns T
Hi. computedAsync returns T for now, but must return T | undefined.
This simple code will throw error in template, because its initial value is undefined, but not `TestResponse`:

```ts
type TestResponse = {
  results: string[];
};

protected readonly test = computedAsync(() => this.httpClient.get<TestResponse>('/test'));
```

```html
@for (result in test().results) { // <--- error is thrown here
{{ result }}
}
```

JeanMeche:
fixes #243


eneajaho:
@all-contributors please add @JeanMeche for code

eneajaho (approved):

PR #252, opened 2024-01-24
refactor: createEffect() !

e-oz:
* Function, returned by `createEffect()` now accepts `Signal<T> | WritableSignal<T>` as an argument;
* `createEffect()` now has optional configuration argument of type `CreateEffectOptions`. Here you can pass an injector, configure if function should retry on error (true by default), and pass `retry()` configuration options. All fields are optional.
* Documentation of `createEffect()` is updated to reflect the changes.

BREAKING CHANGE: second argument, `injector`, is replaced by an object with fields `injector` and `retryOnError`.

### DIFF
--- a/docs/src/content/docs/utilities/Stores/create-effect.md
+++ b/docs/src/content/docs/utilities/Stores/create-effect.md
@@ -3,7 +3,7 @@ title: createEffect
 description: ngxtension/create-effect
 entryPoint: create-effect
 badge: stable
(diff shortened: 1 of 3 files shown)


nartc:
I don't mind if you don't mind the main branch 😅. I'll review when I get on my laptop

e-oz:
Oops, sorry, I've create a new one already, to follow the guidelines: #253

PR #255, opened 2024-01-24
feat: add requireSync to computedAsync for better typings

### LINKED ISSUE #254: prevent computedAsync with startWith to return undefined
Most of my api calls are in a service that return an observable with the `startWith` operator already added to the pipe. I would expect the computedAsync to give me a sync value, but when I use the value in the template I get an undefined.

Example repo here: https://stackblitz.com/edit/stackblitz-starters-zcanws?file=src%2Fmain.ts

```ts
  data = computedAsync(() => this.api.loadAsyncData(this.id()), {
    initialValue: { data: '', loading: true },
  });
```
with the above `data().loading` in the template works.

```ts
  data = computedAsync(() =>
    this.api
      .loadAsyncData(this.id())
      .pipe(startWith({ data: '', loading: true }))
  );
  ```
  using the above crashes.
  
  Would love to see an option to prevent getting undefined back

### PR DESCRIPTION
**Pull request by eneajaho**
Closes #254

### DIFF
--- a/libs/ngxtension/computed-async/src/computed-async.ts
+++ b/libs/ngxtension/computed-async/src/computed-async.ts
@@ -126,6 +126,12 @@ export function computedAsync<T>(
 	} & ComputedAsyncOptions<T>,
 ): Signal<T>;
 
+// We don't support Promises with requireSync
+export function computedAsync<T>(
+	computation: (previousValue?: T | undefined) => Promise<T>,
+	options: ComputedAsyncOptions<T> & { requireSync: true },
+): never;
+
 // Options with a more specific initial value type.
 export function computedAsync<T, const U extends T>(
 	computation: (previousValue?: T | undefined) => ComputationResult<T>,
@@ -152,8 +158,37 @@ export function computedAsync<T, U = undefined>(
 
 		// sourceValue is a signal that will hold the current value and the state of the value
 		let sourceValue: WritableSignal<State<T | U>>;
+
 		if (options?.requireSync) {
-			sourceValue = signal<State<T | U>>({ kind: StateKind.NoValue });
+			if (options.initialValue !== undefined) {
+				sourceValue = signal<State<T | U>>({
+					kind: StateKind.Value,
+					value: options.initialValue as U,
+				});
+			} else {
+				const newSource = computation(undefined);
+
+				// we don't support promises with requireSync - fix Types to prevent this
+				if (isPromise(newSource)) throw new Error(REQUIRE_SYNC_PROMISE_MESSAGE);
+
+				sourceValue = signal<State<T | U>>({ kind: StateKind.NoValue });
+
+				if (isObservable(newSource)) {
+					newSource
+						.subscribe({
+							next: (value) =>
+								sourceValue.set({ kind: StateKind.Value, value }),
+							error: (error) =>
+								sourceValue.set({ kind: StateKind.Error, error }),
+						})
+						.unsubscribe();
+					if (sourceValue()!.kind === StateKind.NoValue) {
+						throw new Error(REQUIRE_SYNC_ERROR_MESSAGE);
+					}
+				} else {
+					sourceValue.set({ kind: StateKind.Value, value: newSource as T });
+				}
+			}
 		} else {
 			sourceValue = signal<State<T | U>>({
 				kind: StateKind.Value,
@@ -192,18 +227,9 @@ export function computedAsync<T, U = undefined>(
 		);
 
 		const sourceResult = source$.subscribe({
-			next: (value) =>
-				sourceValue.set({
-					kind: StateKind.Value,
-					value,
-				}),
-			error: (error) => {
-				// NOTE: Error should be handled by the user (using catchError or .catch())
-				sourceValue.set({
-					kind: StateKind.Error,
-					error,
-				});
-			},
+			next: (value) => sourceValue.set({ kind: StateKind.Value, value }),
+			// NOTE: Error should be handled by the user (using catchError or .catch())
+			error: (error) => sourceValue.set({ kind: StateKind.Error, error }),
 		});
 
 		destroyRef.onDestroy(() => {
@@ -234,6 +260,7 @@ export function computedAsync<T, U = undefined>(
 	});
 }
 
+const REQUIRE_SYNC_PROMISE_MESSAGE = `Promises cannot be used with requireSync. Pass an initialValue or set requireSync to false.`;
 const REQUIRE_SYNC_ERROR_MESSAGE = `The observable passed to computedAsync() did not emit synchronously. Pass an initialValue or set requireSync to false.`;
 
 function createFlattenObservable<T>(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: libs/ngxtension/computed-async/src/computed-async.ts [hunk 1]
```diff
@@ -72,46 +82,160 @@ interface ComputedAsyncOptions<T> extends CreateComputedOptions<T> {
...
-			options?.initialValue ?? undefined,
-		);
+		// sourceValue is a signal that will hold the current value and the state of the value
+		let sourceValue: WritableSignal<State<T>>;
+
+		if (options?.requireSync && options?.initialValue === undefined) {
+			const initialComputation = computation(undefined);
+
+			// we don't support promises with requireSync and no initialValue
+			if (isPromise(initialComputation))
```
JeanMeche, line 191: I'd use curly braces  if this doesn't fit as a oneliner

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -72,46 +82,160 @@ interface ComputedAsyncOptions<T> extends CreateComputedOptions<T> {
...
+
+			sourceValue = signal<State<T>>({ kind: StateKind.NoValue });
+
+			if (isObservable(initialComputation)) {
+				initialComputation
+					.subscribe({
+						next: (value) => sourceValue.set({ kind: StateKind.Value, value }),
+						error: (error) => sourceValue.set({ kind: StateKind.Error, error }),
+					})
+					.unsubscribe();
```
JeanMeche, line 202: This pattern feels a bit weird, maybe add a comment on why this.

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -122,31 +246,46 @@ export function computedAsync<T>(
...
-		});
+				error: (error) => sourceValue.set({ kind: StateKind.Error, error }),
+			});
 
 		destroyRef.onDestroy(() => {
 			effectRef.destroy();
 			sourceResult.unsubscribe();
 		});
 
+		if (options?.requireSync && sourceValue()!.kind === StateKind.NoValue) {
```
JeanMeche, line 263:
```suggestion
		if (options?.requireSync && sourceValue().kind === StateKind.NoValue) {
```
Looks like the assertion isn't required

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -72,46 +96,157 @@ interface ComputedAsyncOptions<T> extends CreateComputedOptions<T> {
...
+				initialComputation
+					.subscribe({
+						next: (value) => sourceValue.set({ kind: StateKind.Value, value }),
+						error: (error) => sourceValue.set({ kind: StateKind.Error, error }),
+					})
+					// we need to unsubscribe because we don't want to keep the subscription
+					// we only care about the initial value
+					.unsubscribe();
+
+				if (sourceValue()!.kind === StateKind.NoValue)
```
nartc, line 215: suggestion: no assertion needed
```suggestion
				if (sourceValue().kind === StateKind.NoValue)
```

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -72,46 +96,157 @@ interface ComputedAsyncOptions<T> extends CreateComputedOptions<T> {
...
+export function computedAsync<T>(
+	computation: ObservableComputation<T>,
+	options: OptionsWithInitialValue<T>,
+): Signal<T>;
+
+export function computedAsync<T>(
+	computation: (
+		previousValue?: T | undefined,
+	) => Promise<T> | Observable<T> | T | undefined,
+	options?: any,
```
nartc, line 184: nit: I would personally set this to be an empty object.
```suggestion
	options: any = {}
```

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -72,46 +96,157 @@ interface ComputedAsyncOptions<T> extends CreateComputedOptions<T> {
...
 		const effectRef = effect(
 			() => {
 				// we need to have an untracked() here because we don't want to register the sourceValue as a dependency
 				// otherwise, we would have an infinite loop
-				const currentValue = untracked(() => sourceValue());
+				const currentState = untracked(() => sourceValue());
+				const currentValue =
+					currentState.kind === StateKind.Value
+						? currentState.value
+						: undefined;
```
nartc, line 238: nit: can be merged
```suggestion
				const currentValue = untracked(() => {
				  const { kind, value } = sourceValue();
				  return kind === StateKind.Value ? value : undefined;
				});
```

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -72,46 +96,157 @@ interface ComputedAsyncOptions<T> extends CreateComputedOptions<T> {
...
 					// we untrack the source$.next() so that we don't register other signals as dependencies
 					untracked(() => sourceEvent$.next(newSource));
+				} else {
+					// if the new source is not an observable or a promise, we set the value immediately
+					untracked(() =>
+						sourceValue.set({ kind: StateKind.Value, value: newSource as T }),
+					);
 				}
 			},
 			{ injector: options?.injector },
```
nartc, line 252: nit: this is not needed because `assertInjector` ensures this `effect()` runs in the provided Injector if it was provided

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -122,31 +257,48 @@ export function computedAsync<T>(
...
+				next: (value) => sourceValue.set({ kind: StateKind.Value, value }),
 				// NOTE: Error should be handled by the user (using catchError or .catch())
-				sourceValue.set(error);
-			},
-		});
+				error: (error) => sourceValue.set({ kind: StateKind.Error, error }),
+			});
 
 		destroyRef.onDestroy(() => {
 			effectRef.destroy();
```
nartc, line 270: nit: feel like `effectRef.destroy()` isn't needed

libs/ngxtension/computed-async/src/computed-async.ts
```diff
@@ -122,31 +257,48 @@ export function computedAsync<T>(
...
-				const value: T = sourceValue()!;
-				return value;
+				const state = sourceValue();
+				switch (state.kind) {
+					case StateKind.Value:
+						return state.value;
+					case StateKind.Error:
+						throw state.error;
+					case StateKind.NoValue:
+						throw new Error(REQUIRE_SYNC_ERROR_MESSAGE);
```
nartc, line 289: question: why do we throw `REQUIRE_SYNC` error here without checking for `options.requiredSync`?

libs/ngxtension/computed-async/src/computed-async.ts
eneajaho: I added it before that, so removing it now.

libs/ngxtension/computed-async/src/computed-async.ts
nartc: Can you add some comments for that case then?

nartc (approved):

PR #259, opened 2024-01-28
feat(merge-from): add mergeFrom

### LINKED ISSUE #221: Merge signals
Looked through the docs of this wonderful library and couldn't find a  signal utility for merging two or more signal (like [RxJS merge](https://rxjs.dev/api/operators/merge)).

Have I overlooked an easy way to do this without an helper function or is this suitable as an ngxtension util?

nartc:
Closes #221

### DIFF
--- a/docs/src/content/docs/utilities/Signals/merge-from.md
+++ b/docs/src/content/docs/utilities/Signals/merge-from.md
@@ -1,0 +1,42 @@
+---
+title: mergeFrom
+description: ngxtension/merge-from
+entryPoint: merge-from
+badge: stable
+contributors: ['chau-tran']
+---
+
+`mergeFrom` is a helper function that merges the values of `Observable`s or `Signal`s and emits the latest emitted value.
+It also gives us the possibility to change the emitted value before emitting it using RxJS operators.
+
+It is similar to `merge()`, but it also takes `Signals` into consideration.
+
+From `ngxtension` perspective, `mergeFrom` is similar to [`computedFrom`](./computed-from.md), but it doesn't emit the combined value, but the latest emitted value by using the `merge` operator instead of `combineLatest`.
+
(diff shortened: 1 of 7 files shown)


libs/ngxtension/merge-from/src/lib/merge-from.ts
```diff
@@ -0,0 +1,114 @@
...
+		if (!isObservable(source)) {
+			source = from(source);
+		}
+
+		return source.pipe(distinctUntilChanged());
+	});
+
+	const merged = operator
+		? merge(...normalizedSources).pipe(operator)
+		: merge(...normalizedSources);
```
LcsGa, line 65: You can simplify this with the `identity` operator from rxjs
```suggestion
	const merged = merge(...normalizedSources).pipe(operator ?? identity);
```

LcsGa: Or you could even remove `?? identity` if `identity` is the default value when you destrucure the result from `parseArgs`

nartc: Thanks

docs/src/content/docs/utilities/Signals/merge-from.md
```diff
@@ -0,0 +1,42 @@
...
+console.log(merged()); // 2
+```
+
+It can be used in multiple ways:
+
+1. Merge multiple `Signal`s
+2. Merge multiple `Observable`s
+3. Merge multiple `Signal`s and `Observable`s
+4. Using initialValue param
+5. Use it outside of an injection context
```
eneajaho, line 42: Should we add a TODO here and open an issue to showcase these ways with examples?

eneajaho (approved):

PR #269, opened 2024-02-07
feat(connect): allow updating partial values with signal through ConnectedSignal

joshuamorony:
This PR proposes two changes:
* Allow supplying a signal to `.with` as well as to `connect` directly
* Allow for partial state updates with an origin signal

The primary use case I am aiming to support here is this (essentially allowing signals driven changes as well as observables):

```ts
	// sources
	private dogs = computedAsync(() => {
		this.someService.version(); // retrigger when version changes
		return this.dbService.getDogs();
	});

	private cats = computedAsync(() => {
		this.someService.version();
		return this.dbService.getCats();
	});

	// centralised state
	private state = signal({
		dogs: undefined,
		cats: undefined,
        })

	constructor() {
		// reducers
		connect(this.state)
                   .with(() => ({ dogs: this.dogs() }))
                   .with(() => ({ cats: this.cats() }));
	}
```

The current behaviour with origin signals does not allow partial updates, this PR changes that, but I may be missing some reason why this was the case in the first place.

If it isn't desirable to include the partial updates change as well, I can update this PR to just include the change which allows this:

```ts
connect(this.state).with(() => ({ dogs: this.dogs() }));
```

as well as this (bottom is currently supported, top isn't):

```ts
connect(this.state, () => ({ dogs: this.dogs() }));
```

### DIFF
--- a/libs/ngxtension/connect/src/connect.spec.ts
+++ b/libs/ngxtension/connect/src/connect.spec.ts
@@ -55,7 +55,7 @@ describe(connect.name, () => {
 			});
 		});
 
-		it.only('should allow connecting with a signal', () => {
+		it('should allow connecting with a partial value from signal', () => {
 			const state = signal({
 				user: {
 					firstName: 'chau',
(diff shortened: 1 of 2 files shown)


eneajaho:
I think that's a great extension of the function. 

I'd like to have both in connect and with function. 

Also, we need to write better docs for all these use cases.

eneajaho (approved):

PR #273, opened 2024-02-11
fix(if-validator): exclude individual validators from composition to allow presence checks

### LINKED ISSUE #262: ifValidator / ifAsyncValidator
Hello,

If I add some ifValidator 

```¨
 'foo': [
        '',
        ifValidator(() => this.areAdditionalFieldsRequired, [Validators.required]),
      ],
```

```
control.hasValidator(Validators.required)
```

Is always `false` even If it's correctly applied

Should I check it somehow differently or would be please possible to add some feature to check .hasIfValidator ? 

Of course I know I can discover required status from the property `this.areAdditionalFieldsRequired`, but I have generic components for different form controls and there it is infered from 

```
isRequired = signal(false);

this.isRequired.set(this.control.hasValidator(Validators.required))
```

So I could do that with additional Input property but it would be nice to infer it from hasValidator function directly

Thank you very much :)

rlmestre:
Closes #262

### DIFF
--- a/libs/ngxtension/if-validator/src/if-validator.spec.ts
+++ b/libs/ngxtension/if-validator/src/if-validator.spec.ts
@@ -1,0 +1,134 @@
+import { FormControl } from '@angular/forms';
+import { of } from 'rxjs';
+import { ifAsyncValidator, ifValidator } from './if-validator';
+
+describe('ifValidator', () => {
+	it('should return null if the condition is false', () => {
+		const control = new FormControl();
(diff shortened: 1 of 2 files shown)


eneajaho:
Hello @rlmestre 
Can you rebase this, and then I think we can merge 🙌

rlmestre:
Done 🙂 thanks @eneajaho !

eneajaho (approved):

PR #275, opened 2024-02-13
fix(map-array): make it work with either writable or readonly source arrays

### LINKED ISSUE #274: mapArray is typed in a way that doesn't work with readonly source arrays
Because `mapArray`'s implementation expects a mutable array type

```ts
export const mapArray = <T, R>(mapFn: (item: T, index: number) => R) =>
  map((array: T[]) => array.map((item, index) => mapFn(item, index)));
```

using it with a readonly source array will make TypeScript complain with [TS2345](https://typescript.tv/errors/#ts2345):

```ts
const source: readonly number[] = [1, 2, 3];

const source$ = of(source).pipe(
  mapArray((item) => item * 2)
  // ^ The type 'readonly number[]' is 'readonly' and cannot be assigned to the mutable type 'number[]'.ts(2345)
);
```

I would argue that any function consuming an array should most likely type the argument as readonly so both mutable and readonly arrays are accepted. Except, of course, if said function is supposed to actually modify the consumed array.

diegovilar:
This PR fixes #274

### DIFF
--- a/libs/ngxtension/map-array/src/map-array.ts
+++ b/libs/ngxtension/map-array/src/map-array.ts
@@ -1,4 +1,4 @@
 import { map } from 'rxjs';
 
 export const mapArray = <T, R>(mapFn: (item: T, index: number) => R) =>
-	map((array: T[]) => array.map((item, index) => mapFn(item, index)));
+	map((array: readonly T[]) => array.map((item, index) => mapFn(item, index)));


eneajaho:
Hello @diegovilar 
Can you also write a test for this? So, we can make sure we don't break anyone? 
Thanks.

nartc:
Closing as there's no activity.

Issue #276, opened 2024-02-14
[RFC] Represent notify/refresh events with signal utility


joshuamorony:
We can run into situations where we want to imperatively indicate that something has happened in order to trigger a re-computation of a signal or an effect. This is the type of event that might have typically been represented by `Subject<void>`.

The core idea for this came originally from @alxhub who recommended, on the removal of `mutate` from signals, that a "version" signal could be used to indicate that a map has been mutated - essentially just incrementing a signal by one.

This proposal deals with the map mutation situation, but also generally any situation where you want to trigger something when something else happens. The usage would look something like this:

```ts
// create the notifier
myTrigger = notify();

// trigger the notifier
myTrigger.notify();
```

```ts
  // reacting to the notifier
  myThing = computedAsync(async () => {
    this.myTrigger() // signal referenced to trigger computation
    // do stuff async stuff and return
  });

  // or in an effect
  constructor(){
    effect(() => {
      this.myTrigger();
      // do something
    });
  }
```

The implementation for this utility that I created looks like this:

```ts
import { Signal, computed, signal } from '@angular/core';

export function notify(): Signal<number> & {
  notify: () => void;
  init: Signal<boolean>;
} {
  const sourceSignal = signal(0);
  const readonlySignal = sourceSignal.asReadonly();

  Object.defineProperties(readonlySignal, {
    notify: {
      value: () => sourceSignal.update((v) => v + 1),
    },
    init: {
      value: computed(() => readonlySignal() === 0),
    },
  });

  return readonlySignal as Signal<number> & {
    notify: () => void;
    init: Signal<boolean>;
  };
}
```

**NOTE:** I have also added an `init` signal so that if you only want to react to the notifier being explicitly triggered (not on init as well) then you can reference the init signal and check that it is `false`.

@eneajaho also created a similar utility that he has been using:

```ts
function triggerer() {
  const s = signal(0);
  Object.defineProperties(readonlySignal, {
    tick: {  value: () => s.update((v) => v + 1) }
  });
  return s as Signal<number> & {tick: () => void};
}
```

Feel free to share any thoughts you have on how this should be implemented and also feel free to leave your thoughts on the naming, for example do you prefer something like:
* `notify` or `notifier` that is triggered with `.notify()`
* `triggerer` that is triggered with `.tick` or `.trigger()`
* or something else

renatoaraujoc:
Hello,

[...]

PR #277, opened 2024-02-14
feat(create-notifier): add createNotifier

joshuamorony:
This is an initial implementation for the `createNotifier` feature discussed here: https://github.com/nartc/ngxtension-platform/issues/276

This incorporates some of the feedback in the discussion and we can continue to iterate on the exact API (personally I am happy with it as is but happy to change as well)

Once the exact implementation is settled I will add some docs to this PR (along with some more tests too)

### DIFF
--- a/libs/ngxtension/create-notifier/src/create-notifier.ts
+++ b/libs/ngxtension/create-notifier/src/create-notifier.ts
@@ -1,0 +1,12 @@
+import { signal } from '@angular/core';
+
+export function createNotifier() {
+	const sourceSignal = signal(0);
+
+	return {
+		notify: () => {
+			sourceSignal.update((v) => v + 1);
+		},
+		listen: sourceSignal.asReadonly(),
+	};
+}
--- a/docs/src/content/docs/utilities/Signals/create-notifier.md
+++ b/docs/src/content/docs/utilities/Signals/create-notifier.md
@@ -1,0 +1,64 @@
+---
+title: createNotifier
+description: ngxtension/create-notifier
+entryPoint: create-notifier
+badge: experimental
+contributors: ['josh-morony']
+---
+
+`createNotifier` provides a way to manually trigger signal re-computations by
+referencing the created notifier signal. Common use cases for this include:
+
+- Manually triggering a "refresh" and reacting to it
+- Mutating a `Map` and having a way to react to that map changing
+- Triggering a re-computation on some kind of event/life cycle hook
+
+It simply creates a standard `signal` that has its value incremented by `1`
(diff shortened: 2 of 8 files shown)


libs/ngxtension/create-notifier/src/create-notifier.ts
```diff
@@ -0,0 +1,12 @@
+import { signal } from '@angular/core';
+
+export function createNotifier() {
```
renatoaraujoc, line 3: `createNotifier()` is fine, but I'd give more context to the name like I mentioned on the discussion, like: 

`createManualSignalNotifier()` or just `createSignalNotifier()`.

Just a matter of preference though, `createNotifier()` is also fine.

libs/ngxtension/create-notifier/src/create-notifier.ts
```diff
@@ -0,0 +1,12 @@
...
+	const sourceSignal = signal(0);
```
renatoaraujoc, line 4: Could think about something like, maybe:

```typescript
const sourceSignal = signal<'FIRST_EMISSION' | number>('FIRST_EMISSION'); // or 'INIT'?

// Then on the notify() updater
return {
  notify: () => {
	  sourceSignal.update((v) => v === 'FIRST_EMISSION' ? 0 : (v+1)); // narrowing ftw :D
  },
  listen: sourceSignal.asReadonly(),
}

// In the effect, if we want to skip the first emission, the code is more readable this way and tells exactly what happened
effect(() => {
  if(myTrigger.listen() === 'FIRST_EMISSION') {
    return;
  }
});

```

libs/ngxtension/create-notifier/src/create-notifier.ts
joshuamorony: Happy to see what more people think but personally I find `createManualSignalNotifier()` too verbose - and I suppose if you do prefer being more explicit in the code you could still make the notifier names themselves more verbose e.g: `myManualSignalNotifier = createNotifier()`

libs/ngxtension/create-notifier/src/create-notifier.ts
joshuamorony: I also prefer the less verbose option here but it's a closer call for me here, in this case `if(myTrigger.listen())` is a bit more mysterious. Again happy to wait and see what the general consensus is on this one, I don't mind it but I would probably prefer a slightly different approach. Maybe even having a separate computed signal e.g:

```ts
if(myTrigger.listenAfterInit()){
  // do stuff
}
```

eneajaho (approved):
Love it!

PR #281, opened 2024-02-15
fix(computed-previous): Don't return null for initial value, but retu…

eneajaho:
…rn the current value of the source signal instead

Is this a breaking change? 🤔


nartc:
Definitely a breaking change

nartc (approved):

PR #282, opened 2024-02-15
feat(create-resource): Initial implementation of createResource

eneajaho:
This PR implements `createResource` and supports `Promise`s only. 

Open for discussion. 

Highly inspired from https://docs.solidjs.com/reference/basic-reactivity/create-resource

Depends on: https://github.com/nartc/ngxtension-platform/pull/281 for `computedPrevious` change.

### DIFF
--- a/libs/ngxtension/create-resource/src/create-resource.ts
+++ b/libs/ngxtension/create-resource/src/create-resource.ts
@@ -1,0 +1,285 @@
+import {
+	Injector,
+	computed,
+	effect,
+	isSignal,
+	signal,
+	untracked,
+	type Signal,
+} from '@angular/core';
+import { assertInjector } from 'ngxtension/assert-injector';
+import { computedPrevious } from 'ngxtension/computed-previous';
+
+interface CreateResourceOptions {
+	injector?: Injector;
+}
+
+export type CreateResourceSource<TValue> = Signal<TValue | undefined>;
+
+export type CreateResourceStatus =
+	| 'unresolved'
+	| 'pending'
+	| 'ready'
+	| 'refreshing'
+	| 'errored';
+
+interface Resource<TValue, TError = string> {
+	data: Signal<TValue>;
+	status: Signal<CreateResourceStatus>;
+	error: Signal<TError | undefined>;
+	loading: Signal<boolean>;
+	latest: Signal<TValue>;
+	refetch: () => void;
+	mutate: (value: TValue) => void;
+	destroy: () => void;
+}
+
+// TODO: FIX types as they don't play well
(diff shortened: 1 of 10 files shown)


libs/ngxtension/create-resource/src/create-resource.ts
```diff
@@ -0,0 +1,285 @@
...
+
+			// we don't want to track anything else except the source and the fetcher
+			untracked(() => {
+				// TODO: do we want to cancel the current promise if it's still pending? it's easy with observables 😅
+				load(promise!);
+			});
+		});
+
+		function load(p: Promise<TValue> | undefined) {
+			if (p && isPromise(p)) {
```
JeanMeche, line 176:
```suggestion
			if (isPromise(p)) {
```

libs/ngxtension/create-resource/src/create-resource.ts
```diff
@@ -0,0 +1,285 @@
...
+				if (state() === 'pending' || state() === 'refreshing') return;
+
+				// if the trigger has changed, we want to refetch and set the state to refreshing
+				if (trigger() !== previousTrigger()) {
+					state.set('refreshing');
+				} else {
+					state.set('pending');
+				}
```
JeanMeche, line 184:
```suggestion

				state.set(trigger() !== previousTrigger() ? 'refreshing' : 'pending');

```

libs/ngxtension/create-resource/src/create-resource.ts
```diff
@@ -0,0 +1,285 @@
...
+	const { source, fetcher, options } = parseArgs<TValue>(args);
+
+	if (fetcher === undefined) {
+		throw new Error('fetcher is required');
+	}
+
+	const value = signal<TValue | undefined>(options.initialValue);
+	const error = signal<TError | undefined>(undefined);
+	const trigger = signal(0);
+	const state = signal<CreateResourceStatus>(
```
ilirbeqirii, line 154: maybe this can be renamed to 'status' thus matching the Type Parameter itself too?

OmerGronich:
This is exactly what I was looking for! I spent the weekend implementing my own version of this (😅) and was introduced to this PR by @tomer953.

Quick question: How is `resource.latest()` different from simply calling `resource()`? I know in SolidJS it doesn't trigger transitions and Suspense, but we don't have that in Angular afaik. 🤔

tomer953:
Oh I need this... yesterday!
Awesome work.

I wonder why exactly its working with Promise and not Observables?
- Observable are way more common in Angular
- Most of the http calls is made by the HttpClient Module who return observable (also, applying interceptors)
-  wrapping with `lastValueFrom(this.http.get('https://api.example.com/data'));` will be extra boilerplate
- built in cancellation

PR #288, opened 2024-02-23
refactor: use Proxy in toObservableSignal() instead of methods linking, to be compatible with RxJS v8

e-oz:
This change removes a bit dirty method of combining the functionality of Angular Signal and RxJS Observable.
Previously, every field of Observable instance was copied to a Signal object.
Now we return a Proxy object, that only returns a method of the Observable instance if the Signal instance does not have the requested method.


eneajaho:
Hello @e-oz 
Can you rebase this ? And then I think we can merge 🙌

e-oz:
Hi @eneajaho 
Done ✅

eneajaho (approved):

Issue #245, opened 2024-01-23
utility function proposal: syncWithLocalStorage


zip-fa:
Hi. I have little utility-function to sync my signal properties with local storage, which probably could be added to ngextension.

local-storage.ts:

```ts
import { Injectable } from '@angular/core';

@Injectable({
  providedIn: 'root'
})
export class LocalStorage {
  get isSupported(): boolean {
    return typeof localStorage !== "undefined";
  }

  setItem<T>(key: string, value: T): void {
    if (!this.isSupported) {
      return;
    }

    try {
      localStorage.setItem(key, JSON.stringify(value));
    } catch (error) {
      if (
        error &&
        ((error as any).name === "QuotaExceededError" ||
          (error as any).name === "NS_ERROR_DOM_QUOTA_REACHED")
      ) {
        console.error(`The ${key} value exceeds the browser storage quota`);
      } else {
        console.error(
          `Error ocurred while serializing the ${key} store value, value not updated`
        );
      }
    }
  }

  getItem<T>(key: string): Partial<T> | null {
    if (!this.isSupported) {
      return null;
    }

    const storedValue = localStorage.getItem(key);

    if (storedValue === "undefined" || storedValue === null) {
      return null;
    }

    try {
      return JSON.parse(storedValue) as T;
    } catch {
      return null;
    }
  }
}

```

sync-with-local-storage.ts:

```ts
export function syncWithLocalStorage(storageKey: string, keysToSignals: Record<string, WritableSignal<any>>): void {
  assertInInjectionContext(syncWithLocalStorage);

  const localStorage = inject(LocalStorage);

  if (!localStorage.isSupported) {
    return null;
  }

  const data = localStorage.getItem<Record<string, any>>(storageKey);

  if (data) {
    for (const key in keysToSignals) {
      const value = data[key];

      if (value !== undefined) {
        keysToSignals[key].set(value);
      }
    }
  }

  effect(() => {
    const storageData: Record<string, WritableSignal<any>> = {};

    for (const key in keysToSignals) {
      storageData[key] = keysToSignals[key]();
    }

    localStorage.setItem(storageKey, storageData);
  });
}
```

usage:

```ts
test = signal<string>('test'); // initial value; will be overwriten with value from localStorage

// call it from here
private localStorageSync = syncWithLocalStorage('testNamespace', {
  test: this.test
});

// or from constructor
constructor() {
  syncWithLocalStorage('testNamespace', {
    test: this.test
  });
}
```

[...]

PR #295, opened 2024-03-01
feat(useLocalStorage): Initial implementation of useLocalStorage 

palexcast:
This addresses https://github.com/nartc/ngxtension-platform/issues/245

Adds a utility that synchronizes localStorage changes across tabs, and has an optional validate method to verify the validity of the object.

I can create a PR for a similar/same thing for sessionStorage if this seems like a good idea/implementation.
Would love some feedback before I do that though :)

### DIFF
--- a/libs/ngxtension/use-local-storage/README.md
+++ b/libs/ngxtension/use-local-storage/README.md
@@ -1,0 +1,3 @@
+# ngxtension/use-local-storage
+
+Secondary entry point of `ngxtension`. It can be used by importing from `ngxtension/use-local-storage`.
--- a/libs/ngxtension/use-local-storage/ng-package.json
+++ b/libs/ngxtension/use-local-storage/ng-package.json
@@ -1,0 +1,5 @@
+{
+	"lib": {
+		"entryFile": "src/index.ts"
+	}
+}
--- a/libs/ngxtension/use-local-storage/project.json
+++ b/libs/ngxtension/use-local-storage/project.json
@@ -1,0 +1,27 @@
+{
+	"name": "ngxtension/use-local-storage",
+	"$schema": "../../../node_modules/nx/schemas/project-schema.json",
+	"projectType": "library",
+	"sourceRoot": "libs/ngxtension/use-local-storage/src",
+	"targets": {
+		"test": {
+			"executor": "@nx/jest:jest",
+			"outputs": ["{workspaceRoot}/coverage/{projectRoot}"],
+			"options": {
+				"jestConfig": "libs/ngxtension/jest.config.ts",
+				"testPathPattern": ["use-local-storage"],
+				"passWithNoTests": true
+			},
+			"configurations": {
+				"ci": {
+					"ci": true,
+					"codeCoverage": true
+				}
+			}
+		},
+		"lint": {
+			"executor": "@nx/eslint:lint",
+			"outputs": ["{options.outputFile}"]
+		}
+	}
+}
--- a/libs/ngxtension/use-local-storage/src/index.ts
+++ b/libs/ngxtension/use-local-storage/src/index.ts
@@ -1,0 +1,1 @@
+export * from './use-local-storage';
--- a/libs/ngxtension/use-local-storage/src/use-local-storage.spec.ts
+++ b/libs/ngxtension/use-local-storage/src/use-local-storage.spec.ts
@@ -1,0 +1,157 @@
+import { TestBed } from '@angular/core/testing';
+import { Subject } from 'rxjs';
+import { useLocalStorage } from './use-local-storage';
+
+describe('useLocalStorage', () => {
+	const key = 'testKey';
+	let setItemSpy: jest.SpyInstance;
+	let getItemSpy: jest.SpyInstance;
+	let storageEventSubject: Subject<Event>;
+
+	beforeEach(() => {
+		storageEventSubject = new Subject<Event>();
+		setItemSpy = jest.spyOn(Storage.prototype, 'setItem');
+		getItemSpy = jest.spyOn(Storage.prototype, 'getItem').mockReturnValue(null); // Default mock to return null
+	});
+
+	afterEach(() => {
+		jest.clearAllMocks();
+	});
+
+	describe('with primitive', () => {
+		it('should set a value in localStorage', () => {
+			TestBed.runInInjectionContext(() => {
+				let { set } = useLocalStorage<string>(key);
+				const testValue = 'value';
+				set(testValue);
+				expect(setItemSpy).toHaveBeenCalledWith(key, JSON.stringify(testValue));
+			});
+		});
+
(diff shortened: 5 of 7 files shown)


ilirbeqirii:
Should this utility function also, follow the "inject***" naming pattern like other functions? @eneajaho

palexcast:
> Should this utility function also, follow the "inject***" naming pattern like other functions? @eneajaho

I've got no strong opinion on the naming of this, I can fix it if you want it to change :) 

I also have some stuff I want to clean up, like passing the validate func as a part of an optional "options" param, as there might be other stuff to add, and it fits the "signal" way of passing options.

eneajaho:
Hello @palexcast 
Thanks for the PR.

Wanted to link this other implementation here https://twitter.com/_wallstreetdev/status/1767693809503084873 

Would like to start a conversation about these implementations and find the best way to go forward.

guzmanoj:
Hello WSD here :)
Disclaimer: I just liked a lot [Mike Hartington's](https://twitter.com/mhartington/status/1765132445895274635) idea but my code is also amateurish. I'm just starting with signals :)
In terms of the implementation it really depends on the direction you guys want to take. The approach I've shared is heavily inspired by React [use-local-storage-state](https://www.npmjs.com/package/use-local-storage-state) and is 100% based on Signals. It also lacks of In-Memory-Data support and proper error handling. This is all I have so far:

```
import { effect, signal, WritableSignal } from '@angular/core';

export interface LocalStorageOptions<T> {
  defaultValue?: T | (() => T);
  storageSync?: boolean;
  serializer?: {
    stringify: (value: unknown) => string;
    parse: (value: string) => unknown;
  };
}

function isFunction(value: unknown): value is (...args: unknown[]) => unknown {
  return typeof value === 'function';
}

function goodTry<T>(tryFn: () => T): T | undefined {
  try {
    return tryFn();
  } catch {
    return undefined;
  }
}

function parseJSON(value: string): unknown {
  return value === 'undefined' ? undefined : JSON.parse(value);
}

export function localStorageSignal<T>(key: string, options?: LocalStorageOptions<T>): WritableSignal<T | undefined> {
  const defaultValue = isFunction(options?.defaultValue) ? options.defaultValue() : options?.defaultValue;
  const serializer = options?.serializer || {
    stringify: JSON.stringify,
    parse: parseJSON
  };
  const storageSync = options?.storageSync ?? true;
  const initialStoredValue = goodTry(() => localStorage.getItem(key));
  const initialValue = initialStoredValue ? goodTry(() => serializer.parse(initialStoredValue) as T) : defaultValue;
  const internalSignal = signal<T | undefined>(initialValue);

  effect(() => {
    const value = internalSignal();
    if (value === undefined) {
      goodTry(() => localStorage.removeItem(key));
    } else {
      goodTry(() => localStorage.setItem(key, serializer.stringify(value)));
    }
  });

  if (storageSync) {
    const onStorage = (event: StorageEvent) => {
      if (event.storageArea === localStorage && event.key === key) {
        const newValue = event.newValue !== null ? (serializer.parse(event.newValue) as T) : undefined;
        internalSignal.set(newValue);
      }
    };
    window.addEventListener('storage', onStorage);
    effect(onCleanup => {
      onCleanup(() => window.removeEventListener('storage', onStorage));
    });
  }

  return internalSignal;
}

```

palexcast:
I'll take a look at both those implementations and see if I can adjust some of my implementation with some input from those.

If there is any more input/wishes regarding the direction of the implementation that would be greatly appreciated :)

palexcast:
@guzmanoj So I finally had some time, and the more I look at your implementation I really like it. It's clean and functional. 
I really don't feel comfortable just copy-pasting  it and taking credit for it though, the credit should be yours ;)

For now I copy-pasted your solution, and renamed it to `injectLocalStorage` like @ilirbeqirii mentioned, as it does adhere more to the convention of the lib.

In regards to error handling, I'm not sure about what the "signal" convention is regarding this.
Should `goodTry` throw an error, or just return `undefined` as the current implementation does.
Maybe @eneajaho has some thoughts?

My current thought regarding this, is that the `parse` function can be used to handle validation/errors. 
One of the tests demonstrate this.

I also unwrapped the serializer so you don't need to pass both `parse` and `stringify`. Adding docs instead should be enough to explain that they act as serializers. (just my opinion).

I've also adjusted the tests, but want to add some more, would be nice to get some feedback before moving on though.

libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
```diff
@@ -0,0 +1,69 @@
+import { effect, signal, type WritableSignal } from '@angular/core';
+
+export type LocalStorageOptions<T> = {
+	defaultValue?: T | (() => T);
```
eneajaho, line 4: Can we add some comments for each option here? 

It would make it easier when passing these options and to read some kind of comments on what they do and how they modify the way this function works.

eneajaho:
Can you also update the PR description, so it is the same as the updated code we have?

libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
```diff
@@ -0,0 +1,69 @@
...
+	const defaultValue = isFunction(options?.defaultValue)
+		? options.defaultValue()
+		: options?.defaultValue;
+	const stringify = isFunction(options?.stringify)
+		? options?.stringify
+		: JSON.stringify;
+	const parse = isFunction(options?.parse) ? options?.parse : parseJSON;
+
+	const storageSync = options?.storageSync ?? true;
+	const initialStoredValue = goodTry(() => localStorage.getItem(key));
```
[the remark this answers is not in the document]
palexcast: Fixed, nice catch! Didn't think about this :D

libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
palexcast: I agree, added some comments :)

palexcast:
@eneajaho Updated the PR in regards to your comments :D

libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
```diff
@@ -0,0 +1,113 @@
...
+	} catch {
+		return undefined;
+	}
+}
+
+function parseJSON(value: string): unknown {
+	return value === 'undefined' ? undefined : JSON.parse(value);
+}
+
+export const injectLocalStorage = <T>(
```
nartc, line 67: nit: please use `assertInjector()` to ensure this function is being called in an Injection Context. Also, can we go with top-level `function` declaration instead of Arrow Fn expression?

eneajaho:
Thanks for this @palexcast 🙌

palexcast:
Thanks for the reviews and help getting this merged @eneajaho :D 
First 'real' contribution to OS, so greatly appreciated!

PR #301, opened 2024-03-09
feat(convert-outputs): add migrations for new output() function

eneajaho:
Added migration support for: 

- EventEmitters -> output
- Observable/Subject/BehaviorSubject -> outputFromObservable

+ support for aliases 
+ docs

### DIFF
--- a/libs/plugin/src/generators/convert-outputs/generator.ts
+++ b/libs/plugin/src/generators/convert-outputs/generator.ts
@@ -120,7 +120,7 @@ function getOutputInitializer(
 			// there is no type
 		} else {
 			const genericTypeOnEmitter = initializer.match(/EventEmitter<(.+)>/);
-			if (genericTypeOnEmitter.length) {
+			if (genericTypeOnEmitter?.length) {
 				type = genericTypeOnEmitter[1];
 			}
 		}
--- a/libs/plugin/src/generators/convert-outputs/generator.ts
+++ b/libs/plugin/src/generators/convert-outputs/generator.ts
@@ -1,0 +1,274 @@
+import {
+	Tree,
+	formatFiles,
+	getProjects,
+	logger,
+	readJson,
+	readProjectConfiguration,
+	visitNotIgnoredFiles,
+} from '@nx/devkit';
+import { readFileSync } from 'node:fs';
+import { exit } from 'node:process';
+import {
+	CodeBlockWriter,
+	Decorator,
+	Node,
+	Project,
+	WriterFunction,
+} from 'ts-morph';
+import { ConvertOutputsGeneratorSchema } from './schema';
+
+class ContentsStore {
+	private _project: Project = null!;
+
+	collection: Array<{ path: string; content: string }> = [];
+
+	get project() {
+		if (!this._project) {
+			this._project = new Project({ useInMemoryFileSystem: true });
+		}
+
+		return this._project;
+	}
+
+	track(path: string, content: string) {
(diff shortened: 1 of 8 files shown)


eneajaho:
cc @joshuamorony  you may like this too 😄

libs/plugin/src/generators/convert-outputs/generator.ts
```diff
@@ -0,0 +1,316 @@
...
+				needsOutputFromObservableImport: true,
+			};
+		}
+	} else {
+		let type = '';
+		if (initializer.includes('EventEmitter()')) {
+			// there is no type
+		} else {
+			const genericTypeOnEmitter = initializer.match(/EventEmitter<(.+)>/);
+			if (genericTypeOnEmitter.length) {
```
kbrilla, line 123: needs to be `if (genericTypeOnEmitter?.length) {`

as `genericTypeOnEmitter` can be undefined 
with this migration run locally

eneajaho: Thanks. On it.

libs/plugin/src/generators/convert-outputs/generator.ts
```diff
@@ -0,0 +1,316 @@
...
+							outputDecorator.remove();
+
+							newProperty.setHasExclamationToken(false);
+
+							newProperty.addJsDoc(
+								'TODO(migration): you may want to convert this to a normal output',
+							);
+						} else {
+							node.replaceWithText(newProperty.print());
+							// remove old class property Output
```
kbrilla, line 293: after `node.replaceWithText(newProperty.print());` 
this code is needed for something like 'this.someOutput.next() to be replaced with this.someOutput.emit()' as this was actually possible on EventEmiter

```
newProperty.findReferencesAsNodes().forEach((reference) => {
								try {
									//we are looking for something like 'this.someOutput.next() as we want to replace it with this.someOutput.emit()'
									//for each identifier we go someOutput, this.someOutput , then DotNotation , then Indentifier and we check that this identifier is 'next'
									const nextIdentifier = reference
										.getParentIfKindOrThrow(SyntaxKind.PropertyAccessExpression)
										.getNextSiblingIfKindOrThrow(SyntaxKind.DotToken)
										.getNextSiblingIfKindOrThrow(SyntaxKind.Identifier);
									if (nextIdentifier.getText() === 'next') {
										if (
											nextIdentifier
												.getParent()
												.getParent()
												.isKind(SyntaxKind.CallExpression)
										) {
											console.log(
												nextIdentifier.getParent().getParent().getText(),
											);
											console.log('\nFOUND next - replacng with emit');
											nextIdentifier.replaceWithText('emit');
										}
									}
								} catch (err) {}
							});
```


I have large project that I'm migrating so I will be adding coments like this if you are ok with it with next things that I stumbled on

now working on types as EventEmitter() is typed EventEmitter<any> but output() is OutputEmitterRef<void>, also found usages of accessing valueChanges on eventEmitters etc so will take care of those as well

kbrilla: of course here simple regex could be used but I feel like this is much safer in terms of migration

eneajaho: Hello @kbrilla 
Thanks for all this 🙌🙏
Would you mind send a PR that we can merge into this PRs branch? 
And then we can merge this PR into main and cut a release?

nartc (approved):

PR #319, opened 2024-04-05
fix(click-outside): remove memory leak

kkachniarz220:


libs/ngxtension/click-outside/src/click-outside.ts
```diff
@@ -50,13 +48,11 @@ export class ClickOutside implements OnInit {
 	ngOnInit() {
 		this.documentClick$
 			.pipe(
-				takeUntil(this.destroy$),
```
eneajaho, line 53: How does removing takeUntil remove a memory leak? 
I don't understand this tbh. 

We want to stop listening to documentClick on directive destroy, the injectDocumentClick is registered in root scope, so that won't ever be unsubscribed afaik.

libs/ngxtension/click-outside/src/click-outside.ts
```diff
@@ -34,14 +37,16 @@ const [injectDocumentClick] = createInjectionToken(() => {
  * <div (clickOutside)="close()"></div>
  *
  */
-@Directive({ selector: '[clickOutside]', standalone: true })
+@Directive({
+	selector: '[clickOutside]',
+	standalone: true,
+	providers: [provideDocumentClick()],
```
eneajaho, line 43: This will create a listener for each directive, while the previous method, would create only one listener for the whole document, and each directive will use the same listener, which makes it more performant I think.

kkachniarz220: How can it be more performant if the observer is active all the time? :) I thought about it and figured that using multiple outsideClicks on one view would be rare.

Is memory leak something we should accept?

Issue #313, opened 2024-03-28
feat: repeat pipe for new @for syntax


Dafnik:
The current implementation of the [Repeat](https://ngxtension.netlify.app/utilities/directives/repeat/) Directive has limited usage for the new template syntax `for` loop. (also it's build on inheritenance instead of composition)

My proposal to circumvent these issues would be a `RepeatPipe`.
It could look something like this:
```ts
@Pipe({
  standalone: true,
  name: 'repeat',
})
export class RepeatPipe implements PipeTransform {
  transform(length: number, startAt = 0): number[] {
    if (Number.isNaN(length) || !Number.isInteger(length) || length < 1) {
      throw new Error(
        `[Repeat] repeat requires an positive integer but ${length} is passed in`
      );
    }
    return Array.from({ length }, (_, index) => index + startAt);
  }
}
```

where the usage is something like this:

```ts

@Component({
  selector: 'app-root',
  standalone: true,
  imports: [RepeatPipe],
  template: `
    @for (i of 10 | repeat; track i) {
      <p>Nr. {{i}}</p>
    }

    <hr/>

    @for (i of 10 | repeat: 10; track i) {
      <p>Nr. {{i}}</p>
    }
  `,
})
export class App { }
```

will result in:
![image](https://github.com/nartc/ngxtension-platform/assets/16242839/c44af84a-1539-45d7-aed8-7d1e4d24b58b)

eneajaho:
That's a good idea.

Dafnik:
Great! I looked into the possibility of providing a migration from the current repeat directive to this new repeat pipe but after looking into how Angular does the new control flow syntax migration I'm not sure about that anymore.

The migration can't be done with a simple regex as it would need to actually check if NgFor (or the CommonModule) is already available for the component.

Do you think we can get away with not providing an automatic migration for this (as it's not easily feasible)?

kkachniarz220:
Migration will not be necessary if the old solution will not be removed. The library can provide two approaches (`*ngFor` and `@for`

PR #320, opened 2024-04-08
feat(repeat-pipe): initial version

Dafnik:
Adds the repeat pipe as described in #313


eneajaho:
Can we add a note in the Repeat directive to let the devs know that a repeat pipe also exists? 

This way, if they want to use the new control flow, they may want to use the repeat pipe instead of the directive.

PR #322, opened 2024-04-09
docs: contributor badges fixes

Dafnik:
Before:
before
before2

After:
after
after2

Also add the missing website property to the contributors collection.

Feel free to dismiss :)


eneajaho (approved):
Love it!

PR #326, opened 2024-04-11
feat: added inject migration

eneajaho:
It supports everything 🎉

Before: 
test (1)

After: 
test (2)

Also inspiration taken from: https://github.com/kreuzerk/inject-pilot/tree/main 

cc @kreuzerk @ilirbeqirii

### DIFF
--- a/libs/plugin/src/generators/convert-di-to-inject/generator.ts
+++ b/libs/plugin/src/generators/convert-di-to-inject/generator.ts
@@ -125,6 +125,16 @@ export async function convertDiToInjectGenerator(
 					.some((namedImport) => namedImport.getName() === 'inject'),
 		);
 
+		const hasHostAttributeTokenImport = sourceFile.getImportDeclaration(
+			(importDecl) =>
+				importDecl.getModuleSpecifierValue() === '@angular/core' &&
+				importDecl
+					.getNamedImports()
+					.some(
+						(namedImport) => namedImport.getName() === 'HostAttributeToken',
+					),
+		);
+
 		const classes = sourceFile.getClasses();
 
 		for (const targetClass of classes) {
@@ -136,6 +146,7 @@ export async function convertDiToInjectGenerator(
 			if (!applicableDecorator) continue;
 
 			const convertedDeps = new Set<string>();
+			let includeHostAttributeToken = false;
 
 			targetClass.getConstructors().forEach((constructor) => {
 				constructor.getParameters().forEach((param, index) => {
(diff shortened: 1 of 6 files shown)


libs/plugin/src/generators/convert-di-to-inject/generator.ts
```diff
@@ -0,0 +1,244 @@
...
+
+				constructor.getParameters().forEach((param) => {
+					if (convertedDeps.has(param.getName())) {
+						param.remove();
+					}
+				});
+
+				if (
+					constructor.getParameters().length === 0 &&
+					constructor.getBodyText().trim() === ''
```
scr2em, line 225: @eneajaho  I think you should check if the constructor has `super()` only, if that's the case the constructor should be removed as well 
```
constructor.getBodyText().trim() === 'super()'
```

ilirbeqirii: I think we should check first if the schematic converts also the DI injections in the plain classes - ones being decorated anyway. if yes, then we can also remove the constructors with super() only. I guess 👀

k3nsei:
Should MyService2 not be public property in class instance? As it’s just argument which should be just in constructor scope.

```typescript
constructor() {
  const myService2 = inject(MyService2);
}
```

eneajaho:
> Shouldn’t MyService2 be just const in constructor scope and not public property in class instance?
> 
> ```ts
> constructor() {
>   const myService2 = inject(MyService2);
> }
> ```

Yeah, you're right about this one. 

Now I'm not sure on what I should do with this. Should we unify the deps or keep it inside the constructor only for that one case?

ilirbeqirii:
> > Shouldn’t MyService2 be just const in constructor scope and not public property in class instance?
> > ```ts
> > constructor() {
> >   const myService2 = inject(MyService2);
> > }
> > ```
> 
> Yeah, you're right about this one.
> 
> Now I'm not sure on what I should do with this. Should we unify the deps or keep it inside the constructor only for that one case?

man, I think you should keep it inside the constructor as that's its scope of work.

eneajaho:
> > > Shouldn’t MyService2 be just const in constructor scope and not public property in class instance?
> > > ```ts
> > > constructor() {
> > >   const myService2 = inject(MyService2);
> > > }
> > > ```
> > 
> > 
> > Yeah, you're right about this one.
> > Now I'm not sure on what I should do with this. Should we unify the deps or keep it inside the constructor only for that one case?
> 
> man, I think you should keep it inside the constructor as that's its scope of work.

Most of the time I've seen this be used as: 

```ts
class Cmp {
  myService2: MyService2;

  constructor(myService2: MyService2;) {
     // just re-assign to a class field
     this.myService2 = myService2;
  }

}
```

So, I'm keep the same behavior by creating it as a variable inside the constructor.

PR #336, opened 2024-04-21
feat: add provider function for configuring route injection functions

eneajaho:
PS: I'm not so sure about this! 

One other way to fix this is to use snapshot and get rid of requireSync at all. 

cc @nartc @fiorelozere @ajitzero


nartc (approved):
LGTM

eneajaho:
Going with a different approach: 

https://github.com/nartc/ngxtension-platform/pull/339

This should not fail anymore, but Ionic users will have to check for the value if it's nullish before using them.

Issue #312, opened 2024-03-28
feat: migration outputs enhancements


eneajaho:
- remove Output from imports if not used anymore
- append output to existing import from angular/core
- migrate the following case correctly
```ts
@Output() someOutput: EventEmitter<DataInterface> = new EventEmitter();

// currently
someOutput = output();

// should be
someOutput = output<DataInterface>();
```

More ideas are welcome 🤗

andrei4ik1997:
Hi @eneajaho, I encountered the following problem during migration. In my application I used:

`@Output() private readonly testChange = new EventEmitter<any>();`

After migration I got

`private readonly testChange = output<any>();`

But when I tried to run the application I received the following error: 

**Cannot use "output" on a class member that is declared as private angular.**

The error was easily resolved **by replacing private with protected**

Maybe it’s worth considering this case in migration?

eneajaho:
Hi, yeah that would be great to be added. Thanks for sharing.

PR #337, opened 2024-04-21
feat(output-migration): enhancements

eneajaho:
- [x] remove `Output` and `EventEmitter` from imports if not used anymore
- [x] append `output`/`outputFromObservable` to existing import from angular/core
- [x] replace `private` with `protected` for outputs to not break existing private scope usages
- [x] migrate the following case correctly
```typescript
@Output() someOutput: EventEmitter<DataInterface> = new EventEmitter();

// currently
someOutput = output();

// should be
someOutput = output<DataInterface>();
```

Closes https://github.com/nartc/ngxtension-platform/issues/312


eneajaho:
In the end of the migration add a link to a discussion inside GitHub so devs can tell us if everything went great, or what can be improved or if there is any bug.

PR #363, opened 2024-05-02
feat(signal-slice): add "Updated" signal for action sources

joshuamorony:
NOTE: This depends on https://github.com/ngxtension/ngxtension-platform/pull/361 being merged, this is branched off of that feature

This is a replacement for the removal of the `actionEffect` API, instead it allows triggering side effects as a result of an `actionSource` emitting via a notifier/version signal (same general concept we used for `createNotifier`).

It adds a `<actionSourceName>Updated` signal for each action source e.g:

```ts
  state = signalSlice({
    actionSources: {
      load: (_state, $: Observable<void>) => $.pipe(
        switchMap(() => this.someService.load()),
        map(data => ({ someProperty: data })
      )
    }
  })

  effect(() => {
    // triggered when `load` emits/completes and on init
    console.log(state.loadUpdated())
  })

  effect(() => {
	if (state.loadUpdated()) {
	  // triggered ONLY when `load` emits/completes
	  // NOT on init
	}
  });
```

This is a much simpler API and avoids the typing issues (described in: https://github.com/ngxtension/ngxtension-platform/pull/361)

### DIFF
--- a/docs/src/content/docs/utilities/Signals/signal-slice.md
+++ b/docs/src/content/docs/utilities/Signals/signal-slice.md
@@ -169,6 +169,38 @@ this subject as an `actionSource` allows you to still trigger it through
 can be accessed on the state object, even if you need to create an external
 subject.
 
+## Action Updates
+
+Each `actionSource` will have an equivalent `Updated` version signal automatically generated that will be incremented each time the `actionSource` emits or completes, e.g:
+
+```ts
+  state = signalSlice({
+    initialState: this.initialState,
+    actionSources: {
+      load: (_state, $: Observable<void>) => $.pipe(
+        switchMap(() => this.someService.load()),
+        map(data => ({ someProperty: data })
+      )
+    }
+  })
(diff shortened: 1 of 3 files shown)


michael-small:
Sounds like a fair tradeoff to me. If the pattern of this PR and the base PR is to de-couple `signalSlice` from formal effect blocks, then having some sort distinction between other `effect` in a file unrelated to the slice and `effect` related to the slice is a nice tradeoff.

eneajaho:
Hello @joshuamorony 
Can you add the breaking change in the commit description if this is a breaking change?

joshuamorony:
@eneajaho we ended up merging #361 and this is ready for review now (no breaking changes in this PR)

eneajaho:
@joshuamorony can you rebase? Then I can merge

joshuamorony:
@eneajaho done

eneajaho (approved):

PR #370, opened 2024-05-11
fix(plugin): add and remove this keyword correctly

### LINKED ISSUE #367: Invalid output that can't compile for convert-di-to-inject with local and "public" dependencies
👋 thanks for this great library. Last week I've used the schematic to migrate our codebase to move away from constructors to `inject`. I stumbled across this edge case, resulting in a project that couldn't be compiled anymore after running the `convert-di-to-inject` migration.

If a dependency is used within the constructor AND outside of it, the `this.` keyword isn't prefixed within the constructor. If the dependency is only used within the constructor this isn't a problem as the dependency will be created as a variable within the constructor.

The following output already shows this behavior (`myService.doSomething();` needs to be `this.myService.doSomething();`)

https://github.com/ngxtension/ngxtension-platform/blob/16b66e514dcff6fd440f12d799d44a6e098f9ee6/libs/plugin/src/generators/convert-di-to-inject/__snapshots__/generator.spec.ts.snap#L48C1-L81C6

This can be fixed by moving the commented code to the `else` block if the dependency is used outside the ctor.

https://github.com/ngxtension/ngxtension-platform/blob/16b66e514dcff6fd440f12d799d44a6e098f9ee6/libs/plugin/src/generators/convert-di-to-inject/generator.ts#L211C1-L230C13

If you want I can create a PR for this.

timdeschryver:
Closes #367

### DIFF
--- a/libs/plugin/src/generators/convert-di-to-inject/generator.ts
+++ b/libs/plugin/src/generators/convert-di-to-inject/generator.ts
@@ -196,18 +196,28 @@ export async function convertDiToInjectGenerator(
 						// check if the service was used inside the constructor with 'this.' prefix
 						// if so, remove the 'this.' prefix and use the service variable directly
 						constructor.getStatements().forEach((statement) => {
-							if (Node.isExpressionStatement(statement)) {
-								const expression = statement.getExpression();
-								if (Node.isCallExpression(expression)) {
-									const expressionText = expression.getText();
-									if (expressionText.includes(`this.${name.toString()}`)) {
-										const newExpression = expressionText.replace(
-											`this.${name.toString()}`,
-											name.toString(),
-										);
-										statement.replaceWithText(newExpression);
-									}
-								}
+							if (!Node.isExpressionStatement(statement)) {
+								return;
+							}
+
+							const callExpression = statement.getExpression();
+							if (!Node.isCallExpression(callExpression)) {
+								return;
+							}
+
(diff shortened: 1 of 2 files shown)


libs/plugin/src/generators/convert-di-to-inject/generator.ts
```diff
@@ -201,33 +219,28 @@ export async function convertDiToInjectGenerator(
 								isReadonly || options.includeReadonlyByDefault || false,
 							leadingTrivia: '  ',
 						});
+
+						param
```
timdeschryver, line 223: This can't use the above logic because it `expressionText.includes` can replace the wrong variables.

libs/plugin/src/generators/convert-di-to-inject/generator.ts
```diff
@@ -192,6 +192,24 @@ export async function convertDiToInjectGenerator(
 							declarationKind: VariableDeclarationKind.Const,
 							declarations: [{ name, initializer }],
 						});
+
+						// check if the service was used inside the constructor with 'this.' prefix
+						// if so, remove the 'this.' prefix and use the service variable directly
+						constructor.getStatements().forEach((statement) => {
```
timdeschryver, line 198: Can't use `findReferences` as it will only result in the scoped variables (which are already correct) and not the `this.` properties.

eneajaho (approved):
Thanks for this ❤️

Issue #290, opened 2024-02-28
convert-signal-inputs fails


grosch-intl:
I have this as the last line of my component's class:

```typescript
#isPromiseBoolean = (value: unknown): value is Promise<boolean> => value instanceof Promise
```

When I run your generator, it fails with this:

<pre>
-- Details --
Path: /src/app/button-enabled-dialog/button-enabled-dialog.component.ts
Text: "...ble\n\n    #isPromiseBoolean = (value: unknown): value is Promise<boolean> => value instanceof Promise\n;\n}\n"
Stack: Error: Cannot get the next when at the end of the iterator.
    at AdvancedIterator.next (C:...\node_modules\ts-morph\dist\ts-morph.js:40:19)
    at DefaultParentHandler.handleNode (C:...\node_modules\ts-morph\dist\ts-morph.js:1542:116)
    at ParentFinderReplacementNodeHandler.handleNode (C:...\node_modules\ts-morph\dist\ts-morph.js:1615:37)
    at NodeHandlerHelper.handleForValues (C:...\node_modules\ts-morph\dist\ts-morph.js:1402:21)
    at #handleChildren (C:...\node_modules\ts-morph\dist\ts-morph.js:1471:25)
    at ParentFinderReplacementNodeHandler.handleNode (C:...\node_modules\ts-morph\dist\ts-morph.js:1461:33)
    at ParentFinderReplacementNodeHandler.handleNode (C:...\node_modules\ts-morph\dist\ts-morph.js:1618:19)
    at NodeHandlerHelper.handleForValues (C:...\node_modules\ts-morph\dist\ts-morph.js:1402:21)
    at #handleChildren (C:...\node_modules\ts-morph\dist\ts-morph.js:1471:25)
    at ParentFinderReplacementNodeHandler.handleNode (C:...\node_modules\ts-morph\dist\ts-morph.js:1461:33)
error Command failed with exit code 1.
</pre>

joshuamorony:
Thanks @grosch-intl - the issue is that this line is not terminated with a semicolon, which is valid JS/TS syntax because of automatic semicolon insertion, but this messes with the processing we are doing in ts-morph

grosch-intl:
Yah, I hate semicolons as they just don't add any value :)  Looking forward to your fix.

Are you planning a generator to handle viewChild, contentChild and the other new ones?

joshuamorony:
[...]

PR #371, opened 2024-05-11
fix(plugin): remove additional semicolumns

timdeschryver:
Remove additional semicolumns created by the input generator.
I might be missing something because it seems like this is desired (#290), but I cannot think of/create an invalid example.

If accepted, I'll also update the output generator as this has the same issue.

Also sorry for not creating an issue first 😅 - I can still do this.


eneajaho:
Thank you for creating this @timdeschryver 🙌

Having it also for the output would be lovely ❤️

I don't think there is a need for an issue.

eneajaho (approved):

PR #378, opened 2024-05-20
feat(inject-migration): add option for ES private field support

### LINKED ISSUE #377: [Inject Migration Enhancement]: Add ES Private Field notation support
When migrating from constructor-based to inject-based DI, add a new option, hence enhancement, to replace the TS `private` access modifier with ES `#private` fields.

ilirbeqirii:
Closes #377

### DIFF
--- a/libs/plugin/src/generators/convert-di-to-inject/generator.ts
+++ b/libs/plugin/src/generators/convert-di-to-inject/generator.ts
@@ -9,7 +9,7 @@ import {
 } from '@nx/devkit';
 import { readFileSync } from 'node:fs';
 import { exit } from 'node:process';
-import { Node, VariableDeclarationKind } from 'ts-morph';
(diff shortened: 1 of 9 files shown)


libs/plugin/src/generators/convert-di-to-inject/generator.ts
```diff
@@ -221,10 +221,14 @@ export async function convertDiToInjectGenerator(
...
+						const propertyName = options.useESPrivateFieldNotation
+							? `#${name}`
+							: name;
+
 						targetClass.insertProperty(index, {
-							name,
+							name: propertyName,
 							initializer,
-							scope,
+							scope: options.useESPrivateFieldNotation ? null : scope,
```
eneajaho, line 231: If the field is protected I don't think we should convert it to private. So I'd only convert private properties to use the # and keep the others as is.

PR #393, opened 2024-05-29
feat: add explicitEffect helper function.

JeanMeche:
Build on top of `effect()`, this helper function allows to only watch the producers passed as dependencies.

The effect itself is run as untracked preventing any hidden side effects.

### DIFF
--- a/libs/ngxtension/explicit-effect/README.md
+++ b/libs/ngxtension/explicit-effect/README.md
@@ -1,0 +1,3 @@
+# ngxtension/explicit-effect
+
+Secondary entry point of `ngxtension`. It can be used by importing from `ngxtension/explicit-effect`.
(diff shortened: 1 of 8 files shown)


libs/ngxtension/explicit-effect/README.md
```diff
@@ -0,0 +1,3 @@
+# ngxtension/create-effect
+
+Secondary entry point of `ngxtension`. It can be used by importing from `ngxtension/create-effect`.
```
nartc, line 3:
```suggestion
# ngxtension/explicit-effect

Secondary entry point of `ngxtension`. It can be used by importing from `ngxtension/explicit-effect`.
```

eneajaho (approved):
Love it!

PR #407, opened 2024-06-06
Fix: new logo paths

michael-small:
This is two similar but separate things

- The Press Kit images were 404'ing while deployed. However, for some reason they resolved fine when ran locally. WSL magic, caching, not sure. Anyways, they are now pointed to work locally and also I would assume as deployed.
- The Spanish main page was still pointing at the old logo.


eneajaho (approved):

eneajaho:
Thanks @michael-small 🙌

PR #409, opened 2024-06-07
fix(input-migration): enhance regexp to consider whitespaces when matching against temp…

### LINKED ISSUE #397: Template migration fails when using optional chaining operator in templates
x is an input transformed to signal. 

The template has a "something = x?.y"

expected the template to change to "x()?.y". but it does not migrate.

ilirbeqirii:
…late

Closes #397 (I also tried to work on the case mentioned on the issue, but am not sure I did it right. It would be helpful if we could have a more complete sample of non-working migration for the opened issue.)

When testing for matches in the template, please make sure we consider when the template code is not properly formatted and the interpolation expression has spaces on both sides.


nartc:
Closing this. I suggest to open a new PR if this is still needed.

PR #416, opened 2024-06-12
feat: add migration for self-closing tags

eneajaho:
Angular supports self-closing tags. This means that you can write tags like `<app-component />` instead of `<app-component></app-component>`.
This is a feature that was introduced in Angular 16.

To run the migration you can run: 

```bash
ng g ngxtension:convert-to-self-closing-tag
```

Closes https://github.com/ngxtension/ngxtension-platform/issues/395


tomer953:
FYI, `angular-eslint` has implemented this[ as eslint rule](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/docs/rules/prefer-self-closing-tags.md) with auto fix.
I would consider adding a reference or "tip" in the util docs, if someone wants to add this rule to avoid closing-tags after the migration.

nartc (approved):

PR #431, opened 2024-06-19
feat(explicitEffect): addition of defer parameter; this parameter all…

mauriziocescon:
As part of the discussion related to explicit deps tracking for the `effect` function, @JeanMeche delivered the implementation of `explicitEffect`. 🎉

This is a small PR that complements the initial implementation by adding an optional `defer` parameter in order to skip the initial callback call (by design effects run once). 

Here is a scenario describing its usage: https://github.com/angular/angular/issues/56155#issuecomment-2137756817

Note that the other frameworks have a similar functionality: 
- solid (on + defer): https://www.solidjs.com/tutorial/reactivity_on, 
- vue (watch, lazy by default): https://vuejs.org/guide/essentials/watchers#eager-watchers, 
- MobX (reaction): https://mobx.js.org/reactions.html#reaction 

As part of the PR, I've also reworked a bit the doc (hope you don't mind).


JeanMeche:
With the upcomming changes to effect, I not sure this is a good idea. 

See angular/angular#56501, we'll have root effects and component effects.

mauriziocescon:
Umh... noticed the PR, but I'm not sure I get all the details (probably I'll wait for your blog-post 😅). 

But in few words: same effect function with different behaviours depending on component and (for example) root service, right? I was assuming the nature (always run at least once) was unchanged...

nartc (approved):

mauriziocescon:
Thanks!

PR #432, opened 2024-06-20
fix(signal-input-migration): preserve line endings

Kiskae:
Without this change the text contained in `value` will be shorter than `end - start` if it contains `\r\n`, since all the `\r` characters will be stripped out. This causes the change offset to be wrong for all subsequent changes.


eneajaho (commented):
Thanks for this @Kiskae 🙌

eneajaho (approved):

eneajaho:
@all-contributors please add @Kiskae for code

PR #435, opened 2024-06-26
Add (failing) tests for setting `undefined` as initial value in `derivedFrom`

mattmoos:
This PR adds failing tests for setting `undefined` as initial value in `derivedFrom` as requested in https://github.com/ngxtension/ngxtension-platform/issues/427

I added tests for observable source and promise source, not sure if both are needed. 

Using Object.hasOwn (or Object.prototype.hasOwnProperty) should fix it. 

The inferred type of the created signal is not correct though, which is also the case when setting null as initial value.


nartc:
@all-contributors please add @mattmoos for code

nartc (approved):

PR #450, opened 2024-07-15
Enhance inject route utilities

Dafnik:
Adds the following options to all utilities and therefore aligns them.

- injectParams
	- transform
	- defaultValue
	- injector
- injectQueryParams
	- defaultValue
	- injector
- injectQueryParams.array
	- defaultValue
	- injector
- injectRouteData
	- defaultValue
	- injector
- injectRouteFragment
	- defaultValue

I also took the chance to rename `initialValue` to `defaultValue` (as I think it's a more fitting name).
`injectQueryParams` supports both values, as to not introduce a breaking change.

### DIFF
--- a/docs/src/content/docs/utilities/Injectors/inject-params.md
+++ b/docs/src/content/docs/utilities/Injectors/inject-params.md
@@ -16,7 +16,9 @@ import { injectParams } from 'ngxtension/inject-params';
 
 ## Usage
 
-`injectParams` when is called, returns a signal with the current route params.
+### Get all route params
+
+`injectParams` returns a signal with the current route params.
 
 ```ts
 @Component({
@@ -28,12 +30,25 @@ class TestComponent {
 }
 ```
 
+#### Transform all params
+
+Or, if we want to transform the params, we can pass a function to `injectParams`.
+
+```ts
+@Component()
+class TestComponent {
+	paramsKeys = injectParams((params) => Object.keys(params)); // returns a signal with the keys of the params
+}
+```
+
+### Specific param
+
 If we want to get the value for a specific param, we can pass the name of the param to `injectParams`.
 
 ```ts
 @Component({
 	template: `
-		@if (user()) {
+		@if (user(); as user) {
 			<div>{{ user.name }}</div>
 		} @else {
 			<div>No user!</div>
@@ -50,11 +65,52 @@ class TestComponent {
 }
 ```
 
-Or, if we want to transform the params, we can pass a function to `injectParams`.
+#### Transform
+
+If you want to additional parse the specific param, we can pass a `parse` function.
 
 ```ts
-@Component()
+@Component({
+	template: `
+		@if (user(); as user) {
+			<div>{{ user.name }}</div>
+		} @else {
+			<div>No user!</div>
+		}
+	`,
+})
 class TestComponent {
-	paramsKeys = injectParams((params) => Object.keys(params)); // returns a signal with the keys of the params
+	userId = injectParams('id', { parse: numberAttribute }); // returns a signal with the value of the id param parsed to a number
+
+	user = derivedFrom(
+		[this.userId],
+		switchMap((id) => this.userService.getUser(id).pipe(startWith(null))),
+	);
+}
+```
+
+#### Default value
+
+If we want to use a default value if there is no value, we can pass a `defaultValue`.
+
+```ts
+@Component({
+	template: `
+		@if (angular(); as angular) {
+			<div>{{ angular.name }}</div>
+		} @else {
+			<div>No Angular version found!</div>
+		}
+	`,
+})
+class TestComponent {
+	angularVersion = injectParams('version', { defaultValue: '19' }); // returns a signal with the value of the id param parsed to a number
+
+	angular = derivedFrom(
+		[this.angularVersion],
+		switchMap((version) =>
+			this.angularService.getAngular(version).pipe(startWith(null)),
+		),
+	);
 }
 ```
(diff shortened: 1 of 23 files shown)


nartc:
is this still WIP?

Dafnik:
Yes, docs and tests are missing.

eneajaho:
Hello @Dafnik 
Thanks for this great PR. 

Regarding 
> I also took the chance to rename initialValue to defaultValue (as I think it's a more fitting name).
injectQueryParams supports both values, as to not introduce a breaking change.

I think initialValue is a better fit there because after the initialValue, the params can change however the router needs, and we shouldn't call it default value (as that would mean that we will reapply it again based on some condition, which we don't because we only need it the first time).

Dafnik:
Hi, thank you very much :)

But it is actually a default value. 😅 

Lets say we have the route `test` and use the URL `/test?a=1234`.
If we use `injectQueryParams('a', {defaultValue: '1111'})` we will get the result `1234`.
But now the route changes to `/test`. The query param is null and the default value `1111` will be emitted.

Or did I get this wrong?

Dafnik:
Ping @eneajaho

krisatopi:
This one should be included in the docs too https://timdeschryver.dev/blog/til-paramsinheritancestrategy-to-always-have-access-to-parent-route-info

eneajaho:
Hello @Dafnik 
Sorry for this taking so long..

As you may have already seen, I've published linkedQueryParam, https://ngxtension.netlify.app/utilities/injectors/linked-query-param/

In order for things to be consistent, I'd like to convert the method transform to be called `parse` in order for utilities to align 
For the `initialValue`, as you said, `defaultValue` would be a better fit.

So, all the utils should contain: `defaultValue`, `injector`, `parse` options. 

If you don't have time to finish this one, I can take over anytime. Please let me know, as I'd like to get these features out as they are really needed.   
 
 Happy new year ✨

eneajaho:
Btw, just re-read this one reply https://github.com/ngxtension/ngxtension-platform/issues/495#issuecomment-2348365828 

Looks like I can take over 🙌

Dafnik:
Yeah please take it over!

The parse function sound good for me! 
Also gives it a distinction of Angular's `tranform`.

I've already tried out linkedQueryParams and I love it!! Really great UX and much better than just having injectQueryParams.

Dafnik:
@eneajaho I took the liberty and finished the PR up.

Ready for your review :)

Dafnik:
Ahh yes. Sorry, this slipped my mind.

I've now renamed all new `transform` functions to `parse`.
`injectQueryParams`, `injectQueryParams.array` and `injectRouteFragment` still support `transform` as to not introduce a breaking change. And I've also marked them as deprecated.

Same for `injectQueryParams`'s and `injectQueryParams.array`'s `initialValue` property.

libs/ngxtension/inject-query-params/src/inject-query-params.ts
```diff
@@ -1,31 +1,44 @@
...
+	string | null
+> &
+	DefaultValueOptions<DefaultValueT> &
+	InjectorOptions & {
+		/**
+		 * The initial value to use if the query parameter is not present or undefined.
+		 *
+		 * @deprecated Use `defaultValue` as a replacement.
+		 */
+		initialValue?: DefaultValueT;
```
eneajaho, line 32: We can write a migration schematic when we remove this one. So it's smooth for everyone when they update.

eneajaho (approved):

libs/ngxtension/inject-query-params/src/inject-query-params.ts
Dafnik: So you mean we shouldn't mark it as deprecated?

docs/src/content/docs/utilities/Injectors/inject-params.md
```diff
@@ -50,11 +65,52 @@ class TestComponent {
...
+	`,
+})
 class TestComponent {
-	paramsKeys = injectParams((params) => Object.keys(params)); // returns a signal with the keys of the params
+	userId = injectParams('id', { parse: numberAttribute }); // returns a signal with the value of the id param parsed to a number
+
+	user = derivedFrom(
+		[this.userId],
+		switchMap((id) => this.userService.getUser(id).pipe(startWith(null))),
+	);
```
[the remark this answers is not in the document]
Dafnik: I haven't really worked with `resource` or `rxResource` yet but it's definitely a better example / showcase.

But, because of the reason above, I would be glad if you could do the examples then.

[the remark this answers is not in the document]
eneajaho: I can do them in another PR.

libs/ngxtension/inject-query-params/src/inject-query-params.ts
eneajaho: No marking them as deprecated is fine.

eneajaho:
I'll make sure to get this merged this week

michael-small:
This may be more of a broad question to make a dedicated issue, but this is apparent across all of the route utils current so I will ask here:

Is there a way that the docs could version the pages so that they don't update until the changes are in a live version? Whether that be automatic or manually. For example, I was wondering why I couldn't `parse` with `injectParams` in the latest version as I was following along with the docs.

eneajaho:
> This may be more of a broad question to make a dedicated issue, but this is apparent across all of the route utils currently so I will ask here:
> 
> Is there a way that the docs could version the pages so that they don't update until the changes are in a live version? Whether that be automatic or manually. For example, I was wondering why I couldn't `parse` with `injectParams` in the latest version as I was following along with the docs. Then I realized I just had to jump back to the docs' previous commit's pages.

I forgot to cut a new release 👀

eneajaho:
Should be good now! v4.4.0

PR #459, opened 2024-07-17
feat: add mirror utility

eneajaho:


mauriziocescon:
You can make it even more generic by simply allowing a "reactive func" instead of just a signal! 

Here is an example: https://stackblitz.com/edit/stackblitz-starters-ebanrk?file=src%2Fmain.ts

eneajaho:
> You can make it even more generic by simply allowing a "reactive func" instead of just a signal! This way it's closer to a `writableComputed`.
> 
> Here is an example: https://stackblitz.com/edit/stackblitz-starters-ebanrk?file=src%2Fmain.ts

Yeah that would make it better.

eneajaho:
https://github.com/angular/angular/issues/55673#issuecomment-2288435547

PR #467, opened 2024-07-24
fix: remove duplicate test in signal slice spec

jdegand:
The only thing different is the description.

### DIFF
--- a/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
+++ b/libs/ngxtension/signal-slice/src/signal-slice.spec.ts
@@ -138,11 +138,6 @@ describe(signalSlice.name, () => {
 			state.age();
 			expect(testFn).toHaveBeenCalled();
 		});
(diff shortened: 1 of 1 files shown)


libs/ngxtension/signal-slice/src/signal-slice.spec.ts
```diff
@@ -138,11 +138,6 @@ describe(signalSlice.name, () => {
 			state.age();
 			expect(testFn).toHaveBeenCalled();
 		});
-
-		it('should connect lazy source after signal value is accessed', () => {
-			state().age;
```
ajitzero, line 143: What is this a duplicate of?

The previous one is `state.age()` while this one is `state().age`

jdegand: Yeah. You are right and I should have paused checked again.

PR #490, opened 2024-08-27
Feat: Add tapOnce and tapOnceOnFirstTruthy Operators

endlacer:
This pull request adds two new standalone RxJS operators, tapOnce and tapOnceOnFirstTruthy, to the utility library. These operators allow for conditional execution of functions based on emitted values.

tapOnce: Executes the provided function only once when the value at the specified index is emitted.
tapOnceOnFirstTruthy: Executes the provided function only once when the first truthy value is emitted.

What are your thoughts on these operators? They seem particularly useful for scenarios where you need to initialize something, such as a modal element, as soon as the first value from a source is emitted.

This is my first PR on GitHub, so I appreciate your patience :)! Any contributions, feedback, or change requests are more than welcome.

### DIFF
--- a/docs/src/content/docs/utilities/RxJS-Operators/tap-once.md
+++ b/docs/src/content/docs/utilities/RxJS-Operators/tap-once.md
@@ -1,0 +1,73 @@
+````markdown
+---
+title: tapOnce / tapOnceOnFirstTruthy
+description: Standalone RxJS operators for executing functions conditionally on emitted values.
+entryPoint: ngxtension/tap-once
+badge: stable
+contributors: ['endlacer']
+---
+
+## Import
+
+```typescript
+import { tapOnce, tapOnceOnFirstTruthy } from 'ngxtension/tap-once';
+```
+````
+
+## Usage
+
+### tapOnce
+
+Executes the provided function only once when the value at the specified index is emitted.
+
+```typescript
+import { from } from 'rxjs';
+import { tapOnce } from 'ngxtension/tap-once';
+
+const in$ = from([1, 2, 3, 4, 5]);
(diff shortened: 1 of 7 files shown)


eneajaho:
Hello @endlacer 
Thanks for creating this PR, 
I would like to know more about the usecases for this operator, can you give some examples which this solves. 
Currently it's hard to understand where this would be useful and if other projects can benefit from this.

endlacer:
hi @eneajaho ,
I used the tapOnceOnFirstTruthy operator for initialization of a modal dialog. For example, I have an open$ Subject of type ModalData. The ModalComponent subscribes to an Observable derived from this Subject. When it receives its first truthy value, it performs some initialization before opening the modal in the subscribe block.

Here's a simplified example:
```typescript
import { Subject } from 'rxjs';
import { tapOnceOnFirstTruthy } from './operators';

class ModalComponent {
  private open$ = new Subject<ModalData | null>();

  constructor() {
    this.open$
      .pipe(
        tapOnceOnFirstTruthy((data) => {
          // Perform one-time initialization
          this.initializeModal(data);
        })
      )
      .subscribe((data) => {
        if (data) {
          this.openModal(data);
        } else {
          this.closeModal();
        }
      });
  }

  private initializeModal(data: ModalData) {
    // One-time setup logic
  }

  private openModal(data: ModalData) {
    // Logic to open the modal
  }

  private closeModal() {
    // Logic to close the modal
  }
}
```
This approach ensures that the initialization logic runs only once, even if the modal is opened and closed multiple times.

Another example could be a user filling out a multipage form and when reaching the midpoint you want to log that to some analyze-backend. But you only want to do it once and not on every back/forth navigation. This can be done with a sperate subscription with filter + take(1), but this operator is cleaner

PR #493, opened 2024-09-11
Feature/delete host directive

HamedFathi:
I have written a simple directive to answer [this GitHub issue](https://github.com/angular/angular/issues/18877).


Dafnik:
I've ran into this behaviour many times. Nice solution.

Does this break other host directives applied to a component or directive?

And if so, there should probably be a big warning in the docs.

PR #497, opened 2024-09-13
fix(plugin): support outputs without inline initialization

ShacharHarshuv:
Previously, this script would fail with a cryptic message (`cannot read property of undefined (reading includes)`) for input like:
```ts
@Output() myOutput: EventEmitter<string>;
```

The changes support that use case and:
- Uses the initialization in the constructor to infer the type (and removes it)
- Creates an initialization (`= output()`) if it doesn't exist


nartc (approved):

nartc:
@all-contributors please add @ShacharHarshuv for code

nartc:
@ShacharHarshuv sorry that this falls on you but apparently the lint target fails because `convert-output` is affected by this change. Can you fix the lint issue? Thank you.

ShacharHarshuv:
@nartc I believe it should be fixed now

PR #499, opened 2024-09-15
refactor: Handle getting value from disabled form controls in `form-events` util

### LINKED ISSUE #491: Handle getting value from disabled form controls in `form-events` util
Something occurred to me just now

## Problem
The current state of the `form-events` utils (`allEventsObservable` and `allEventsSignal`, names of them may change in a future version) doesn't handle getting the current value of form controls if a particular one is disabled: https://stackblitz.com/edit/stackblitz-starters-j5jucv?file=src%2Fform-events-utils.ts,src%2Fmain.ts

Example form group

```ts
  fb = inject(NonNullableFormBuilder);
  form = this.fb.group({
    firstName: this.fb.control('', Validators.required),
    lastName: this.fb.control(''),
  });
```

Here is a scenario that the current util's code and the stackblitz demonstrates
```ts
  ngOnInit() {
    this.form.setValue({ firstName: 'Michael', lastName: 'Small' });
    this.form.controls.firstName.disable(); <--- this is the issue
  }

  // Value of "value" given by the utils if `firstName` was disabled, 
  //     since the util currently doesn't use the `.getRawValue()` trick, 
  //     but rather the `ValueChangeEvent`
  // "value": {
  //  "lastName": "Small"
  // },
```

This is because the `value` of a reactive form changes if a control is disabled. However, people like myself included normally use `form.getRawValue()` to get the value of all controls even if they are disabled. And this was my original intent as well, but it got lost as these utils were made, and because the util handles a whole form group being disabled fine but single controls being disabled is the issue I forgot about.

## Proposed Solution
However, with a proposed change, the value would include disabled individual controls like this: https://stackblitz.com/edit/stackblitz-starters-ujq78h?file=src%2Fform-events-utils.ts,src%2Fmain.ts

```ts
  ngOnInit() {
    this.form.setValue({ firstName: 'Michael', lastName: 'Small' });
    this.form.controls.firstName.disable(); <--- this would be handled by the `.getRawValue()` change I am proposing
  }
  // "value": {
  //   "firstName": "Michael",
  //   "lastName": "Small"
  // },
```

This proposed change would replace using the value change events from the new unified form events API and just use the old fashioned `.getRawValue()` mapping like in the example code listed.

## Considerations

I would just make a PR for this now, as my original intent was to account for this. And in my experience, people commonly want to use the raw value trick to always get the value, form control being disabled or not. Here is an example of that discussion that I weighed into which spurred on me making this issue: https://old.reddit.com/r/Angular2/comments/1f9501j/best_way_to_force_nonnullable_form_control_values/llj6f9s/. 

HOWEVER, this may arguably be a breaking change. I see this as a bug from my intent and some people's intent when they want a form's value, but others may see this as a feature. 

However that however: the util as is handles a whole form group being disabled by still giving its value, so that is another weird aspect of what may be expected out of the util. I think that is a point towards making this change however.

Lastly, and probably a huge factor
- The util still doesn't have docs: https://github.com/ngxtension/ngxtension-platform/pull/451
- It will probably be renamed: https://github.com/ngxtension/ngxtension-platform/pull/391#issuecomment-2273165339

So it is still up in the air for some changes and I am guessing not that used. But I don't want to assume.

## Thoughts? I think the proposed solution would be good, but the considerations may be relevant to users

michael-small:
## About

In all my efforts to get this util all together, I overlooked one of the most important things when it comes to getting form values from reactive forms (for what I argue in the issue is what most people want): getting the value of the form and each control even if the form is disabled.

Closes #491: "Handle getting value from disabled form controls in `form-events` util"

## Notes
- naming of `form-events` and its functions is outstanding.
- Two existing tests were misphrased and one was not accounting for both of the utils. I threw in changes to those since I was in the test suite anyways.
- @eneajaho about our conversation on naming the signal util: thoughts on `toFormState`? I feel like it is the compliment to the proposed observable util name of `fromFormEvents`.

### DIFF
--- a/libs/ngxtension/form-events/src/form-events.spec.ts
+++ b/libs/ngxtension/form-events/src/form-events.spec.ts
@@ -115,7 +115,7 @@ describe('Form Events', () => {
 		);
 	});
 
-	it('returns a signal with the value, status, pristine, and touched values of a form after it has been interacted with', async () => {
+	it('returns an observable with the value, status, pristine, and touched values of a form after it has been interacted with', async () => {
 		const fixture: ComponentFixture<FormEventsComponent> =
 			TestBed.configureTestingModule({
 				imports: [FormEventsComponent],
@@ -165,8 +165,28 @@ describe('Form Events', () => {
 
 		fixture.detectChanges();
 
-		const signalVals: HTMLElement = fixture.debugElement.query(
+		const observableVals: HTMLElement = fixture.debugElement.query(
 			By.css('[data-testid="observable-values-initial-value-overwritten"]'),
+		).nativeElement;
+
+		expect(flattenJsonPipeFormatting(observableVals.textContent)).toBe(
+			flattenJsonPipeFormatting(`{
+            "value": {
+                "name": "custom"
+            },
+            "status": "VALID",
+            "touched": false,
+            "pristine": true,
+            "valid": true,
+            "invalid": false,
+            "pending": false,
+            "dirty": false,
+            "untouched": true
+            }`),
+		);
+
+		const signalVals: HTMLElement = fixture.debugElement.query(
+			By.css('[data-testid="signal-values-initial-value-overwritten"]'),
 		).nativeElement;
 
 		expect(flattenJsonPipeFormatting(signalVals.textContent)).toBe(
--- a/libs/ngxtension/form-events/src/form-events.spec.ts
+++ b/libs/ngxtension/form-events/src/form-events.spec.ts
@@ -205,6 +205,63 @@ describe('Form Events', () => {
             }`),
 		);
 	});
+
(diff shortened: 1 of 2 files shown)


nartc:
Can you rebase from latest `main` real quick to see if there's anything wrong with Nx 20 and latest Angular minor?

msmallest:
The tests as well as a quick manual test of the test app seemed to work fine. 

I am terribad at git stuff that isn't commit/push/pull, like rebase, so I may have screwed it up trying to just rebase following a guide. Idk if it doubling those commits up and the merge was intended from what I think I understand about rebases. I can try to revert this stuff, or is it fine as is? I'm pretty sure you are going to squash this anyways.

nartc (approved):

Issue #506, opened 2024-09-26
[enhancement]: better typing for injectLocalStorage


tanishqmanuja:
```
  names = injectLocalStorage<string[]>("names", {
    defaultValue: [],
  });
  // names: WritableSignal<string[] | undefined>
```

Currently even if we supply defaultValue as an option in `injectLocalStorage<T>()` the returned WritableSignal is of type `T | undefined`

Workaround for now,
```
  names = injectLocalStorage("names", {
    defaultValue: [],
  }) as WritableSignal<string[]>;
```

PR #508, opened 2024-09-27
feat: handle defaultValue not undefined typing

Ostromecky:
Link to the issue: [#506](https://github.com/ngxtension/ngxtension-platform/issues/506)

This is my first contribution to an Open Source project—the issue was labeled as a "Good First Issue."

This implementation aims to allow the compiler to differentiate between the provided arguments, ensuring that the correct code block is returned with the appropriate type. I used function overloading to handle different implementations.

My main concern is the interpretation of defaultValue. While it could be used as fallback data for initialValue, the issue is that we also listen for storage events, which might return an undefined value. I'm also unsure about the naming conventions.

I'm grateful for every remark and comment.

### DIFF
--- a/docs/src/content/docs/utilities/Injectors/inject-local-storage.md
+++ b/docs/src/content/docs/utilities/Injectors/inject-local-storage.md
@@ -55,9 +55,7 @@ Options to configure the behavior of the local storage signal.
 Here's a basic example of using `injectLocalStorage`:
 
 ```typescript
-const username = injectLocalStorage<string | undefined>('username', {
-	storageSync: true
-});
+const username = injectLocalStorage<string>('username');
 
 username.set('John Doe');
 username.update((username) => 'Guest ' + username);
@@ -66,13 +64,23 @@ effect(() => {
 	console.log(username());
 });
 // Use `username` in your component to get or set the username stored in local storage.
+// The value might be null or undefined if default value is not provided.
 ```
 
 Fallback value can be provided using the `defaultValue` option:
 
 ```typescript
 const username = injectLocalStorage<string>('username', {
-	defaultValue: 'Guest'
+	defaultValue: 'Guest',
 });
 // If the key 'username' is not present in local storage, the default value 'Guest' will be used.
+```
+
+Storage synchronization can be enabled using the `storageSync` option:
+
+```typescript
+const username = injectLocalStorage<string>('username', {
+	storageSync: true,
+});
+// Changes to the local storage will be reflected across browser tabs.
 ```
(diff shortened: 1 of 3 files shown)


eneajaho:
Hello @Ostromecky 
Thanks for fixing this and congrats on landing your first contribution 🎉

Regarding your question about listening to storage events that may return undefined. 

It would be great if we document this part (in the docs, as it may lead to bugs in the future), and also let's always add undefined in the type if the synchronisation is enabled, this way, the user always will have to check for undefined in that case. 

What do you think?

libs/ngxtension/inject-local-storage/src/inject-local-storage.spec.ts
```diff
@@ -29,6 +29,16 @@ describe('injectLocalStorage', () => {
 			});
 		}));
 
+		it('should return data of defaultValue', () => {
+			TestBed.runInInjectionContext(() => {
+				const defaultValue = 'default';
+				const localStorageSignal = injectLocalStorage<string>(key, {
```
eneajaho, line 35: Can we include a test that only checks for the type and not just the value?

Ostromecky: It can be difficult to differentiate between an exact type union and undefined (e.g., string | undefined). At runtime, we typically just receive a string. I improved the tests to cover cases where a default value is provided (ensuring we don't return undefined) and cases without a default value, where undefined is returned from local storage. Please verify it once again and let me know if it works for you. Additionally, I think those tests document the behavior I explained before. Nevertheless, I'd add appropriate info in docs as well.

docs/src/content/docs/utilities/Injectors/inject-local-storage.md
```diff
@@ -55,9 +55,8 @@ Options to configure the behavior of the local storage signal.
 Here's a basic example of using `injectLocalStorage`:
 
 ```typescript
-const username = injectLocalStorage<string>('username', {
-	defaultValue: 'Anonymous',
-	storageSync: true,
+const username = injectLocalStorage<string | undefined>('username', {
+	storageSync: true
 });
```
eneajaho, line 60: I would move this below to showcase the storageSync method, instead of showcasing it as a basic example. 

A basic example would just be : `const username = injectLocalStorage<string>('username');` and tell the developer that it will also contain the null or undefined (whatever we use as default) value in the type

eneajaho (changes requested):

eneajaho (approved):

eneajaho:
@all-contributors please add @Ostromecky for code

tanishqmanuja:
Thanks @Ostromecky 💯

Ostromecky:
> Thanks @Ostromecky 💯

It was a pleasure to become a part of something bigger :)

Issue #523, opened 2024-11-14
Dispatch storage event on update, for sync signals in the same tab


MillerSvt:
I suggest to dispatch a storage event whenever an update is made. This way, you can keep the signals in sync in the current tab.

https://github.com/ngxtension/ngxtension-platform/blob/323805fb515c47f3fd348e6abb6d30f01d95ab47/libs/ngxtension/inject-local-storage/src/inject-local-storage.ts#L138

_I compare it with my own [implementation](https://stackblitz.com/edit/stackblitz-starters-qlp1ma?file=src%2Finject-storge.ts), but I would like to make it open source_

PR #524, opened 2024-11-14
test(inject-local-storage): fix tests (prepare for #523)

MillerSvt:
Some tests didn't test anything at all.


eneajaho (approved):
Thanks @MillerSvt 🙌

PR #525, opened 2024-11-14
feat(inject-local-storage): dispatch storage event on change (#523)

MillerSvt:
Related to: #523

### DIFF
--- a/libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
+++ b/libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
@@ -1,7 +1,7 @@
+import { DOCUMENT } from '@angular/common';
 import {
 	DestroyRef,
 	InjectionToken,
-	effect,
 	inject,
 	signal,
 	type Injector,
@@ -75,11 +75,11 @@ function isFunction(value: unknown): value is (...args: unknown[]) => unknown {
 	return typeof value === 'function';
 }
 
-function goodTry<T>(tryFn: () => T): T | undefined {
+function goodTry<T>(tryFn: () => T, defaultValue: T): T {
 	try {
 		return tryFn();
 	} catch {
-		return undefined;
+		return defaultValue;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
```diff
@@ -124,29 +124,69 @@ const internalInjectLocalStorage = <R>(
...
 
 		if (storageSync) {
 			const onStorage = (event: StorageEvent) => {
 				if (event.storageArea === localStorage && event.key === key) {
-					const newValue =
-						event.newValue !== null
-							? (parse(event.newValue) as R)
-							: defaultValue;
-					internalSignal.set(newValue);
+					rawValue.set(event.newValue);
```
eneajaho, line 189: This looks like a job for observables to be honest. I'd like to see this part converted to a subject that emits and this way we can handle things easily. We have too many effects at this point inside the utility. We need only one I guess, the others can be replaced by the subject subscription. What do you think?

MillerSvt: We can change both the serialized value (based on the storage change event) and the deserialized value (programmatically in the application). This can create difficulties when using observables, since changes must be propagated in two directions.

To avoid an endless update cycle, I introduced two signals. Since the serialized value is always string or null, the signal will not trigger any effects if it changes to the same value, and the update cycle ends with.

However, I get your idea. We can use one effect by removing the second signal for the serialized value and adding an equivalence check function for the deserialized. I'll be back soon with new solution.

MillerSvt: Ready to review

eneajaho: Hi @MillerSvt 
I gave this some more thought, and I think we can ditch the effect completly. Because local storage is always sync, we can just write directly to it and update it whenever we use set or update. 

Here is the impl I came up with, also I could remove all the Testbed.flushEffects, since the impl doesn't depend on effect anymore. The only thing I didn't test is the sync in the same tab, that one needs to be tested. 

```ts

const internalInjectLocalStorage = <R>(
  key: string,
  options: LocalStorageOptions<R>,
  defaultValue: R,
): WritableSignal<R> => {
  const stringify = isFunction(options.stringify)
    ? options.stringify
    : JSON.stringify;
  const parse = isFunction(options.parse) ? options.parse : parseJSON;
  const storageSync = options.storageSync ?? true;

  function syncValueWithLocalStorage<R>(value: R) {
    const newValue = goodTry(
      () => {
        if (value === undefined || value === null) return null;
        // we stringify the value using the stringify function
        return stringify(value);
      },
      null,
    );

    if (newValue === null) {
      localStorage.removeItem(key);
    } else {
      localStorage.setItem(key, newValue);
    }

    if (storageSync) {
      try {
        // We notify other consumers in this tab about changing the value in the store for synchronization
        window.dispatchEvent(new StorageEvent(`storage`, {key, newValue, storageArea: localStorage}));
      } catch {
        // ignore errors
      }
    }
  }

  return assertInjector(injectLocalStorage, options.injector, () => {
    const localStorage = inject(NGXTENSION_LOCAL_STORAGE);
    const destroyRef = inject(DestroyRef);

    // first we get the initial value from localStorage
    const initialStoredValue = goodTry(() => localStorage.getItem(key), null);

    // then we parse the initial value using the parse function, or the default value if the parsing fails
    const initialSignalValue = initialStoredValue
      ? goodTry(() => parse(initialStoredValue) as R, defaultValue)
      : defaultValue;

    const internalSignal = signal<R>(initialSignalValue);
    const originalSet = internalSignal.set;

    const set = (value: R) => {
      // set the value in the signal using the original set function
      originalSet(value);

      // then we refresh the value in localStorage and notify other consumers in this tab about the change
      syncValueWithLocalStorage(value);
    };

    if (storageSync) {
      const onStorage = (event: StorageEvent) => {
        if (event.storageArea === localStorage && event.key === key) {
          const newValue = event.newValue !== null ? (parse(event.newValue) as R) : defaultValue;
          internalSignal.set(newValue);
        }
      };

      window.addEventListener('storage', onStorage);
      destroyRef.onDestroy(() => {
        window.removeEventListener('storage', onStorage);
      });
    }

    const update = (updateFn: (currentValue: R) => R) => {
      const currentValue = internalSignal();
      set(updateFn(currentValue));
    };

    return Object.assign(internalSignal, { set, update });
  });
};
```

MillerSvt: fixed

libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
```diff
@@ -124,27 +123,74 @@ const internalInjectLocalStorage = <R>(
...
+				}
+
+				if (newValue === null) {
+					localStorage.removeItem(key);
+				} else {
+					localStorage.setItem(key, newValue);
+				}
+
+				// We notify other consumers in this tab about changing the value in the store for synchronization
+				window.dispatchEvent(
```
nartc, line 151: nit: use DI for `window`

nartc (commented):
LGTM. Just one minor comment

libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
nartc: use the DOCUMENT token to access the window object

MillerSvt: fixed

MillerSvt: fixed

nartc (approved):

nartc:
Thank you

PR #526, opened 2024-11-15
feat: add linkedQueryParam implementation, tests and docs

eneajaho:
Added docs for most of the usecases:
 - string
 - number
 - boolean
 - object
 - array

Added example for filtering and pagination with resource API. 

Added multiple tests to test all the possible usecases.

Added 2 helper functions for easier parsing (paramToNumber, paramToBoolean), with support for defaultValue. 

Here is some self-explanatory code examples: 

```ts
export class MyCmp {

        readonly searchQuery = linkedQueryParam('searchQuery');
	
	// Parse a boolean query parameter
	showHidden = linkedQueryParam('showHidden', { parse: paramToBoolean({defaultValue: false}) });

	// Parse a number query parameter
	count = linkedQueryParam('count', { parse: paramToNumber({defaultValue: 0}) });

	// Parse an object query parameter
	filter = linkedQueryParam('filter', {
		parse: (value) => JSON.parse(value ?? '{}'),
		stringify: (value) => JSON.stringify(value),
	});

	// Parse an array query parameter
	selectedCategoriesIds = linkedQueryParam<string[]>('selectedCategoriesIds', {
		parse: (value) => value?.split(',').map((x) => x.trim()) ?? [],
		stringify: (value) => value.join(','),
	});

	selectCategoryId(id: string) {
           // we are working with writable signals, so we can set/update them
           this.selectedCategoriesIds.update((x) => [...x, id]);
	}
}
```

Playground: https://stackblitz.com/edit/stackblitz-starters-v1gtrd?file=src%2Ftodos.ts

### DIFF
--- a/apps/test-app/src/app/linked-query-param/linked-query-param-basic.component.ts
+++ b/apps/test-app/src/app/linked-query-param/linked-query-param-basic.component.ts
@@ -1,0 +1,70 @@
+import { JsonPipe } from '@angular/common';
+import { Component, effect, inject } from '@angular/core';
+import { FormsModule } from '@angular/forms';
+import { Title } from '@angular/platform-browser';
+import { linkedQueryParam } from 'ngxtension/linked-query-param';
+
+@Component({
+	standalone: true,
+	template: `
+		<div>
+			<pre>SearchQuery: {{ searchQuery() | json }}</pre>
+
+			<div>
+				<label>
+					Different inputs same signal:
+					<input [(ngModel)]="searchQuery" name="a" placeholder="searchQuery" />
+					<input [(ngModel)]="searchQuery" name="b" placeholder="searchQuery" />
+				</label>
+			</div>
+
+			<div>
+				<label>
+					Different signals same query param:
+					<input
+						[(ngModel)]="differentSignalWithSearchQuery"
+						name="c"
+						placeholder="differentSignalWithSearchQuery"
+					/>
+				</label>
+			</div>
+
+			<button type="button" (click)="searchQuery.set('cool')">cool</button>
+			<button type="button" (click)="searchQuery.set('great')">great</button>
+			<button type="button" (click)="searchQuery.set(null)">Reset</button>
+		</div>
+	`,
+	imports: [FormsModule, JsonPipe],
+	styles: `
+		div {
+			padding: 20px;
+		}
+	`,
+})
+export default class LinkedQueryParamStringCmp {
+	private titleService = inject(Title);
+
+	searchQuery = linkedQueryParam('searchQuery', {
+		// NOTE in docs that serialize should come after parse in order for types to work correctly
+		// stringify: (x) => `${x}`,
+		// parse: (x) => x ? parseInt(x, 10) : null,
+	});
+
+	differentSignalWithSearchQuery = linkedQueryParam('searchQuery', {
+		defaultValue: 'default',
+	});
+
+	constructor() {
+		effect(() => {
+			const searchQuery = this.searchQuery();
+			console.log('searchQuery Type: ', typeof searchQuery);
+			console.log('searchQuery Value: ', searchQuery);
+
+			if (searchQuery) {
+				this.titleService.setTitle(searchQuery);
+			} else {
+				this.titleService.setTitle('No search query!');
+			}
+		});
+	}
+}
--- a/apps/test-app/src/app/linked-query-param/linked-query-param-insideObject.component.ts
+++ b/apps/test-app/src/app/linked-query-param/linked-query-param-insideObject.component.ts
@@ -1,0 +1,133 @@
+import { JsonPipe } from '@angular/common';
+import { Component, effect, inject } from '@angular/core';
+import { FormsModule } from '@angular/forms';
+import { Router } from '@angular/router';
+import { linkedQueryParam } from 'ngxtension/linked-query-param';
+
+@Component({
+	standalone: true,
+	template: `
+		<div>
+			<label>
+				SearchQuery:
+				<input [(ngModel)]="filterState.searchQuery" name="searchQuery" />
+
+				<button type="button" (click)="filterState.searchQuery.set(null)">
+					reset
+				</button>
+			</label>
+			<br />
+
+			<label>
+				Size:
+				<select [(ngModel)]="filterState.pageSize" name="pageSize">
+					<option value="10">10</option>
+					<option value="20">20</option>
+					<option value="30">30</option>
+					<option [ngValue]="null">null</option>
+				</select>
+			</label>
+			<br />
+			<label>
+				Show deleted:
+				<input
+					[(ngModel)]="filterState.showDeleted"
+					name="showDeleted"
+					type="checkbox"
+				/>
+			</label>
+			<br />
+
+			<label>
+				Page:
+				<input [(ngModel)]="filterState.pageNumber" name="pageNumber" />
+
+				<button type="button" (click)="filterState.pageNumber.set(1)">1</button>
+				<button type="button" (click)="filterState.pageNumber.set(2)">2</button>
+				<button type="button" (click)="filterState.pageNumber.set(3)">3</button>
+				<button type="button" (click)="filterState.pageNumber.set(null)">
+					null
+				</button>
+			</label>
+			<br />
+
+			<label>
+				SortBy:
+				<select [(ngModel)]="filterState.sortBy" name="sortBy">
+					<option value="name">name</option>
+					<option value="age">age</option>
+				</select>
+			</label>
+			<br />
+
+			<label>
+				Direction:
+				<select [(ngModel)]="filterState.direction" name="direction">
+					<option value="asc">asc</option>
+					<option value="desc">desc</option>
+				</select>
+			</label>
+
+			<br />
+
+			<button type="button" (click)="filterState.pageSize.set(10)">
+				pageSize 10
+			</button>
+			<button type="button" (click)="filterState.pageSize.set(20)">
+				pageSize 20
+			</button>
+			<button type="button" (click)="filterState.pageSize.set(null)">
+				pageSize null
+			</button>
+			<button type="button" (click)="resetAll()">reset all</button>
+			<hr />
+			<br />
+			<br />
+			<hr />
+			<br />
+		</div>
+	`,
+	imports: [FormsModule, JsonPipe],
+	styles: `
+		div {
+			padding: 20px;
+		}
+	`,
+})
+export default class LinkedQueryParamInsideObjectCmp {
+	private router = inject(Router);
+
+	filterState = {
+		searchQuery: linkedQueryParam('searchQuery'),
+		showDeleted: linkedQueryParam('showDeleted', {
+			parse: (x) => x === 'true',
+		}),
+		pageNumber: linkedQueryParam('page', { parse: (x) => (x ? +x : null) }),
+		pageSize: linkedQueryParam('pageSize', { parse: (x) => (x ? +x : null) }),
+		sortBy: linkedQueryParam('sortBy', { defaultValue: 'name' }),
+		direction: linkedQueryParam<'asc' | 'desc'>('direction', {
+			defaultValue: 'asc',
+		}),
+	};
+
+	constructor() {
+		Object.values(this.filterState).forEach((x) => {
+			effect(() => {
+				console.log(x());
+			});
+		});
+	}
+
+	resetAll() {
+		this.router.navigate([], {
+			queryParams: {
+				searchQuery: null,
+				showDeleted: null,
+				page: null,
+				pageSize: null,
+				sortBy: null,
+				direction: null,
+			},
+		});
+	}
+}
--- a/libs/ngxtension/linked-query-param/src/linked-query-param.spec.ts
+++ b/libs/ngxtension/linked-query-param/src/linked-query-param.spec.ts
@@ -1,0 +1,508 @@
+import {
+	Component,
+	inject,
+	Injector,
+	numberAttribute,
+	OnInit,
+	signal,
+	WritableSignal,
+} from '@angular/core';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+
+import { ActivatedRoute, provideRouter } from '@angular/router';
+import { RouterTestingHarness } from '@angular/router/testing';
+import {
+	linkedQueryParam,
+	paramToBoolean,
+	paramToNumber,
+} from './linked-query-param';
+
+interface MyParams {
+	a: string;
+	b: number;
+	c: boolean;
+	d: {
+		e: string;
+		f: number;
+		g: boolean;
+	};
+}
+
+@Component({ standalone: true, template: `` })
+export class SearchComponent {
+	route = inject(ActivatedRoute);
+
+	defaultBehavior = linkedQueryParam('defaultBehavior'); // WritableSignal<string | null>
+	defaultBehaviorWithDefault = linkedQueryParam('defaultBehaviorWithDefault', {
+		defaultValue: 'default',
+	}); // WritableSignal<string>
+
+	parseBehavior = linkedQueryParam('parseBehavior', {
+		parse: (x: string | null) => parseInt(x ?? '0', 10),
+	}); // WritableSignal<number>
+
+	parseBehavior1 = linkedQueryParam('parseBehavior', {
+		parse: (x: string | null) => (x ? parseInt(x, 10) : x),
+	}); // WritableSignal<number | string | null>
+
+	parseBehaviorWithParseAndNumberAttribute = linkedQueryParam(
+		'parseBehaviorWithParseAndNumberAttribute',
+		{ parse: numberAttribute },
+	); // WritableSignal<number>
+
+	withBooleanParams = linkedQueryParam('withBooleanParams', {
+		parse: (x) => x === 'true',
+		stringify: (x) => (x ? 'true' : 'false'),
+	}); // WritableSignal<boolean>
+
+	withBuiltInBooleanParamNoDefault = linkedQueryParam('withBooleanNoDefault', {
+		parse: paramToBoolean(),
+	}); // WritableSignal<boolean | null>
+
+	withBuiltInBooleanParamWithDefault = linkedQueryParam(
+		'withBooleanWithDefault',
+		{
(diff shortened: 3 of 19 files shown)


libs/ngxtension/linked-query-param/src/linked-query-param.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+
+		expect(instance.defaultBehaviorWithDefault()).toBe('value'); //THINK: should this be null?
+
+		tick();
+		expect(
+			instance.route.snapshot.queryParams['defaultBehaviorWithDefault'],
+		).toBe(undefined);
+	}));
+
+	//
```
eneajaho, line 185: TODO: also test what happens when we update a query param and some other linkedQueryParam is listening to the same thing, test the timing of that and try to fix it.

eneajaho:
Should I just have two effects instead of the hack? 

```ts

/**
 * Creates a signal that is linked to a query parameter.
 *
 * @param paramKey The name of the query parameter.
 * @param config Configuration options for the signal.
 * @returns A signal that is linked to the query parameter.
 */
export function linkedQueryParam<T = string | null>(
	paramKey: string,
	config?: {
		/**
		 * Function to parse the query param value before it is passed to the signal.
		 * This is useful if the query param is a string and you want to parse it to a number or boolean.
		 */
		parse?: (x: string | null) => T;

		/**
		 * Function to transform the value you pass to the set function.
		 * This is useful if you want to transform the value before it is passed to the query param.
		 */
		transform?: (x: T | null) => string | null;

		/**
		 * The default value to use if the query param is not present.
		 */
		defaultValue?: T;

		/**
		 * The injector to use to inject the router and activated route.
		 */
		injector?: Injector;
	} & Pick<
		NavigationExtras,
		| 'queryParamsHandling'
		| 'onSameUrlNavigation'
		| 'replaceUrl'
		| 'skipLocationChange'
	>,
): WritableSignal<T> {
	const injector = assertInjector(linkedQueryParam, config?.injector);

	const parseFn = config?.parse ?? ((x: string | null) => x as unknown as T);
	const transformFn =
		config?.transform ??
		((x: T | null | undefined) =>
			x as unknown as string | number | boolean | null);

	// @ts-ignore
	return runInInjectionContext(injector, () => {
		const route = inject(ActivatedRoute);
		const coalescedNavigation = inject(CoalescedNavigation);

		const {
			queryParamsHandling,
			onSameUrlNavigation,
			replaceUrl,
			skipLocationChange,
		} = config ?? {};
		coalescedNavigation._navigationExtras = {
			queryParamsHandling,
			onSameUrlNavigation,
			replaceUrl,
			skipLocationChange,
		};

		// get the initial value to use below
		const initialValue =
			parseFn(route.snapshot.queryParams[paramKey]) || config?.defaultValue || undefined;

		// create a signal that is updated whenever the query param changes
		const queryParamValue = toSignal(
			route.queryParams.pipe(
				map((x) => parseFn(x[paramKey]) || config?.defaultValue),
			),
			{ initialValue },
		);


    // TODO: convert this to a linked signal to remove the effect
    // that synchronizes the source signal with the query param value
    // the source signal is what we return to the caller
		const source = signal<T | undefined>(queryParamValue());

		effect(() => {
			const x = queryParamValue();
			// update the source signal whenever the query param changes
			untracked(() => source.set(x));
		});

    effect(() => {
      const sourceValue = source();

      untracked(() => {
        // when the source signal changes, update the query param
        // store the new value in the current keys so that we can coalesce the navigation
        coalescedNavigation._currentKeys[paramKey] = transformFn(sourceValue!);
        // schedule the navigation event
        coalescedNavigation.scheduleNavigation();
      });
    });

		return source;
	});
}

```

libs/ngxtension/linked-query-param/src/linked-query-param.spec.ts
eneajaho: I got this working in the latest version. So adding a test is the only thing left.

apps/test-app/src/app/linked-query-param/linked-query-param-insideObject.component.ts
```diff
@@ -0,0 +1,134 @@
...
+		}
+	`,
+})
+export default class LinkedQueryParamInsideObjectCmp {
+	private router = inject(Router);
+
+	filterState = {
+		searchQuery: linkedQueryParam('searchQuery'),
+		showDeleted: linkedQueryParam('showDeleted', {
+			parse: (x) => x === 'true',
```
JeanMeche, line 103: Should we make that behavior the default for the value `true` (also same for `false`)

apps/test-app/src/app/linked-query-param/linked-query-param-insideObject.component.ts
```diff
@@ -0,0 +1,134 @@
...
+		}),
+		pageNumber: linkedQueryParam('page', { parse: (x) => (x ? +x : null) }),
```
JeanMeche, line 105: Same here we could do the detection automatically with `/\d*/.test(x)`

apps/test-app/src/app/linked-query-param/linked-query-param-insideObject.component.ts
```diff
@@ -0,0 +1,134 @@
...
+		sortBy: linkedQueryParam('sortBy', { defaultValue: 'name' }),
+		direction: linkedQueryParam<'asc' | 'desc'>('direction', {
+			defaultValue: 'asc',
+		}),
+	};
+
+	constructor() {
+		Object.keys(this.filterState).forEach((key) => {
+			effect(() => {
+				// @ts-expect-error TODO: fix this
```
JeanMeche, line 116: You have a todo left here

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+	| 'replaceUrl'
+	| 'skipLocationChange'
+>;
+
+/**
+ * Service to coalesce multiple navigation calls into a single navigation event.
+ */
+@Injectable({ providedIn: 'root' })
+export class LinkedQueryParamGlobalHandler {
+	private router = inject(Router);
```
JeanMeche, line 45:
```suggestion
	private _router = inject(Router);
```

For consistency ?

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+	/**
+	 * @internal
+	 * The navigation extras that will be used on the next navigation event.
+	 */
+	private _navigationExtras: NavigationExtras = {};
+	/**
+	 * @internal
+	 * The notifier that will be used to schedule the navigation event.
+	 */
+	_schedulerNotifier = createNotifier();
```
JeanMeche, line 60:
```suggestion
	private _schedulerNotifier = createNotifier();
```

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+	/**
+	 * Sets the navigation extras that will be used on the next navigation event.
+	 */
+	setCurrentNavigationExtras(config: Partial<NavigateMethodFields>) {
+		const {
+			queryParamsHandling,
+			onSameUrlNavigation,
+			replaceUrl,
+			skipLocationChange,
+		} = config ?? {};
```
JeanMeche, line 89: if `config` is nullable, it should be reflected on the param type.

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+
+		if (queryParamsHandling || queryParamsHandling === '')
+			this._navigationExtras.queryParamsHandling = queryParamsHandling;
+		if (onSameUrlNavigation)
+			this._navigationExtras.onSameUrlNavigation = onSameUrlNavigation;
+		if (replaceUrl) this._navigationExtras.replaceUrl = replaceUrl;
+		if (skipLocationChange)
+			this._navigationExtras.skipLocationChange = skipLocationChange;
```
JeanMeche, line 97: nit: I'd drop the curly only for the one liners

apps/test-app/src/app/linked-query-param/linked-query-param-basic.component.ts
```diff
@@ -0,0 +1,72 @@
...
+			<!--			<div>-->
+			<!--				<label>-->
+			<!--					Different signals same query param:-->
+			<!--					<input-->
+			<!--						[(ngModel)]="differentSignalWithSearchQuery"-->
+			<!--						name="c"-->
+			<!--						placeholder="differentSignalWithSearchQuery"-->
+			<!--					/>-->
+			<!--				</label>-->
+			<!--			</div>-->
```
nartc, line 30: question: intended comment?

apps/test-app/src/app/linked-query-param/linked-query-param-basic.component.ts
```diff
@@ -0,0 +1,72 @@
...
+
+			untracked(() => {
+				const searchQuery = this.searchQuery();
+				if (searchQuery) {
+					this.titleService.setTitle(searchQuery);
+				} else {
+					this.titleService.setTitle('No search query!');
+				}
+			});
+		});
```
nartc, line 70: nit: this seems a bit confusing to me. The `effect` seems to rely on `searchQuery` to update the `title` but the actual `searchQuery` value is in `untracked`? I am aware that the track of `searchQuery()` happens in the `console.log`

apps/test-app/src/app/linked-query-param/linked-query-param-insideObject.component.ts
```diff
@@ -0,0 +1,134 @@
...
+		pageSize: linkedQueryParam('pageSize', { parse: (x) => (x ? +x : null) }),
+		sortBy: linkedQueryParam('sortBy', { defaultValue: 'name' }),
+		direction: linkedQueryParam<'asc' | 'desc'>('direction', {
+			defaultValue: 'asc',
+		}),
+	};
+
+	constructor() {
+		Object.keys(this.filterState).forEach((key) => {
+			effect(() => {
```
nartc, line 115: nit: I wouldn't spawn an `effect` for each `filterState` key

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+	 */
+	_schedulerNotifier = createNotifier();
+
+	constructor() {
+		effect(() => {
+			// listen to the scheduler notifier to schedule the navigation event
+			this._schedulerNotifier.listen();
+
+			// we need to untrack the navigation call in order to not register any other signal as a dependency
+			untracked(() => this.navigate().then());
```
nartc, line 68: nit: my personal style is that if `then()` call is empty, I'd use `void` to annotate that: I skip the return value of this expression call.

```suggestion
			untracked(() => void this.navigate());
```

libs/ngxtension/linked-query-param/src/linked-query-param.ts
nartc: Agree.

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+	 * This will be used on the next navigation event.
+	 */
+	setParamKeyValue(key: string, value: SerializeReturnType) {
+		this._currentKeys[key] = value;
+	}
+
+	/**
+	 * Sets the navigation extras that will be used on the next navigation event.
+	 */
+	setCurrentNavigationExtras(config: Partial<NavigateMethodFields>) {
```
nartc, line 83: nit: if `config` isn't required as inferred by `config ?? {}` below, provide the default value to the parameter

```suggestion
	setCurrentNavigationExtras(config: Partial<NavigateMethodFields> = {}) {
```

libs/ngxtension/linked-query-param/src/linked-query-param.ts
nartc: For cases like this, I'd keep all curly braces for consistency.

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+ *
+ * @param key The name of the query parameter.
+ * @param options Configuration options for the signal.
+ * @returns A signal that is linked to the query parameter.
+ */
+export function linkedQueryParam<T = string>(
+	key: string,
+	options: LinkedQueryParamOptions & {
+		parse: ParseFn<T>;
+		serialize: SerializeFn<T>;
```
nartc, line 171: issue: I see this as an issue because of the semantic of Serialization in general. Pairing `parse` with `serialize` seems odd to me. I suggest `deserialize` and `serialize` even though `deserialize` is longer than `parse`, the semantic is clear as to what Serialization means in this case for query params.

nartc, line 171: question: I see a remark about the order of these two functions to maintain type safety, wouldn't `NoInfer` work?

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+	},
+): WritableSignal<T> & {
+	set: SignalSetFn<T | null>;
+	update: SignalUpdateFn<T | null>;
+};
```
nartc, line 176: nit/suggestion: I would make it explicitly clear that we'll be overriding `set` and `update` to allow `null`

```suggestion
): Omit<WritableSignal<T>, 'set' | 'update'> & {
	set: SignalSetFn<T | null>;
	update: SignalUpdateFn<T | null>;
};
```

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+		serialize?: SerializeFn<T>;
+	},
+): WritableSignal<T> {
+	const injector = assertInjector(linkedQueryParam, options?.injector);
+
+	if (options?.defaultValue !== undefined && options?.parse) {
+		throw new Error(
+			'linkedQueryParam: You cannot have both defaultValue and parse at the same time!',
+		);
+	}
```
nartc, line 270: nit: I would move this error case to the top of the function even before asserting the Injector

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+ * ```
+ *
+ * Will return 1 if the query param is not present in the url.
+ */
+export function paramToNumber(config: {
+	defaultValue: number;
+}): (x: string | null) => number;
+export function paramToNumber(config?: {
+	defaultValue?: number | null | undefined;
+}): (x: string | null) => number | null;
```
nartc, line 369: question: would `export function paramToNumber(): number | null` work? It looks like as soon as I provide a `config` object, the first overload takes over and I can't provide `null | undefined` to `defaultValue` anyway so why would the second overload needs to declare the optional `config?` object at all?

libs/ngxtension/linked-query-param/src/linked-query-param.ts
```diff
@@ -0,0 +1,414 @@
...
+ *
+ * Will return true if the query param is not present in the url.
+ * Otherwise, it will return whatever the query param value is.
+ */
+export function paramToBoolean(config: {
+	defaultValue: boolean;
+}): (x: string | null) => boolean;
+export function paramToBoolean(config?: {
+	defaultValue?: boolean | null | undefined;
+}): (x: string | null) => boolean | null;
```
nartc, line 405: question: same as above about `paramToNumber` overload

nartc (changes requested):

libs/ngxtension/linked-query-param/src/linked-query-param.ts
eneajaho: keeping all with curly braces

apps/test-app/src/app/linked-query-param/linked-query-param-basic.component.ts
eneajaho: forgot about this one. refactored

libs/ngxtension/linked-query-param/src/linked-query-param.ts
eneajaho: I wanted to have the possibility to have the type number only on the paramToNumber if the user provides a number and remove null as a type that the signal can have.

libs/ngxtension/linked-query-param/src/linked-query-param.ts
eneajaho: After private discussion, going with `parse` and `stringify`.

libs/ngxtension/linked-query-param/src/linked-query-param.ts
eneajaho:
```ts
  _withNumberParse = linkedQueryParam('page', { parse: paramToNumber({defaultValue:1}) });

// will produce this type

//                               👇 you will never get null from the type
_withNumberParse: WritableSignal<number> & {   
   set: SignalSetFn<number | null>   
   update: SignalUpdateFn<number | null> 
 }
```

libs/ngxtension/linked-query-param/src/linked-query-param.ts
JeanMeche: Yeah it's exactly the reason why `NoInfer` is used in `resource`.

eneajaho: not sure how to do that. help

libs/ngxtension/linked-query-param/src/linked-query-param.ts
nartc: image
you would still keep your overloads but simplify one of the 2 overloads. This is what I meant

libs/ngxtension/linked-query-param/src/linked-query-param.ts
eneajaho: Keeping it as is for the moment. We can fix it in another PR later ofcourse, if anyone wants to spend some time fixing it.

nartc (approved):

PR #533, opened 2024-12-20
feat: add migration to convert from ngOnDestroy to DestroyRef

michaelbe812:
# DRAFT

Draft for a migration schematic to migrate from `ngOnDestroy` lifecycle hook to `DestroyRef`.

Please bare with me - I have not written many schematics for now :) 

## todos

- [ ] tests
- [ ] docs

## Feedback wanted
I would like to know if you generally like the idea. 

Also I would like to know which style you would prefer:

```ts
// option 1 (currently implemented this way in the schematic)
export class SomeComponent {
  readonly #destroy = inject(DestroyRef).onDestroy(() => {
    // whatever was before in ngOnDestroy
  });
}

// option 2 
export class SomeComponent {
  readonly #destroyRef = inject(DestroyRef)
  
  constructor(){
   this.#destroyRef.onDestroy(() => {
    // whatever was before in ngOnDestroy
  });
}

}
```

Anything else I should consider?

### DIFF
--- a/libs/plugin/generators.json
+++ b/libs/plugin/generators.json
@@ -36,6 +36,11 @@
 			"factory": "./src/generators/convert-to-sfc/generator",
 			"schema": "./src/generators/convert-to-sfc/schema.json",
 			"description": "libs/plugin/src/generators/convert-to-sfc/ generator"
+		},
+		"convert-to-destroy-ref": {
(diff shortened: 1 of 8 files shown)


michaelbe812:
@eneajaho 
just asking for general feedback:
Would you like to see this kind of migration in ngxtension?
If yes, what general style should the migration follow (see pr description)

looking forward :)

PR #537, opened 2024-12-26
feat: add support for v19

eneajaho:


eneajaho:
For the first time since the release, this one would be breaking because we're cutting the angular version to only support v19 and higher, since standalone is true and we removed allowSignalWrites from utilties.

endlacer:
closes https://github.com/ngxtension/ngxtension-platform/issues/536
any updates on this?

nartc:
@eneajaho you would need to amend the commit with a `BREAKING CHANGE` commit body

MillerSvt:
@eneajaho I rebased this branch onto master. And fix versions table. You can cherry-pick your commit from #597.

PR #543, opened 2025-01-19
feat: add injectSignalHistory implementation, tests and docs

eneajaho:
Useful when you need to implement undo/redo functionality on your app

https://github.com/user-attachments/assets/f181353c-a8ff-45f6-a7e3-58e037a642ed

Inspired from https://vueuse.org/core/useRefHistory/

### DIFF
--- a/apps/test-app/src/app/app.component.ts
+++ b/apps/test-app/src/app/app.component.ts
@@ -56,6 +56,10 @@ import { RouterLink, RouterOutlet } from '@angular/router';
 			<li>
 				<a routerLink="/linked-query-param">Linked Query Param</a>
 			</li>
+
+			<li>
+				<a routerLink="/signal-history">Signal History</a>
+			</li>
 		</ul>
 
 		<hr />
--- a/apps/test-app/src/app/app.config.ts
+++ b/apps/test-app/src/app/app.config.ts
@@ -79,6 +79,11 @@ export const appConfig: ApplicationConfig = {
 				loadChildren: () =>
 					import('./linked-query-param/routes').then((x) => x.routes),
 			},
+			{
+				path: 'signal-history',
+				loadComponent: () =>
+					import('./signal-history/signal-history.component'),
+			},
 		]),
 	],
 };
--- a/apps/test-app/src/app/signal-history/signal-history.component.ts
+++ b/apps/test-app/src/app/signal-history/signal-history.component.ts
@@ -1,0 +1,311 @@
+import { JsonPipe } from '@angular/common';
+import { ChangeDetectionStrategy, Component, signal } from '@angular/core';
+import { FormsModule } from '@angular/forms';
+import { signalHistory } from 'ngxtension/signal-history';
+
+@Component({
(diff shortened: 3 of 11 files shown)


Harpush:
The `explicitEffect` looks like a great candidate for the new v19 linked signal

eneajaho:
> Just an idea: you could avoid effect/explicitEffect() usage, because on input you always have a writable signal, so you can just override set() and update().

I started with that, but reverted to an effect in order to allow the users to update the signal however they want during that microtask, so all the changes are scheduled instead of it being fully synchronous.

libs/ngxtension/inject-signal-history/src/inject-signal-history.ts
```diff
@@ -0,0 +1,184 @@
...
+ * @param source The writable signal to track.
+ * @param options Configuration options for the history.
+ * @returns An object with `history`, `undo`, `redo`, `canUndo`, and `canRedo` properties.
+ */
+export function injectSignalHistory<T>(
+	source: WritableSignal<T>,
+	options?: {
+		/**
+		 * The maximum number of history records to store.
+		 * @default 50
```
michael-small, line 36:
```suggestion
		 * @default 100
```

libs/ngxtension/inject-signal-history/src/inject-signal-history.ts
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/**
+ * Enhances a writable signal with undo/redo history functionality.
+ *
+ * @param source The writable signal to track.
+ * @param options Configuration options for the history.
+ * @returns An object with `history`, `undo`, `redo`, `canUndo`, and `canRedo` properties.
+ */
+export function injectSignalHistory<T>(
```
JeanMeche, line 31: I personnaly don't like the "inject" part of the name. 
You don't inject a signal, you create one, hense I would either go with `signalHistory()` or `createSignalHistory()`

libs/ngxtension/inject-signal-history/src/inject-signal-history.ts
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * A signal indicating if undo is possible.
+	 */
+	canUndo: Signal<boolean>;
+	/**
+	 * A signal indicating if redo is possible.
+	 */
+	canRedo: Signal<boolean>;
+} {
+	const capacity = options?.capacity ?? 100; // Default capacity is 100 records
```
JeanMeche, line 76: I'd guard that function with `options?.injector && assertInInjectionContext(...);`

libs/ngxtension/inject-signal-history/src/inject-signal-history.ts
eneajaho: Went with `signalHistory()`

libs/ngxtension/inject-signal-history/src/inject-signal-history.ts
eneajaho: Forgot about it!

PR #555, opened 2025-02-04
fix(ngx-control-value-accessor): accessing properties is outside of a reactive context

### LINKED ISSUE #547: Decourage / remove option to access signal value in NgxControlValueAccessor via get value(){}
### Issue

We noted in our project that if you access the value of ngxControlValueAccessor through the getter (value) that it registers the signal as a dependency. 

I discussed this in my company and team and we think this is something might worth more pointing out or removing. 

In the context of Angular Signals it is unusual to access a Signal without having to explicitly call the signal. Hence one would assume that the signal wouldn't be tracked as a dependency. 

```
	@Input()
	public set value(value: T) {
		this.value$.set(value);
	}

	public get value() {
		return this.value$();
	}
```

https://github.com/ngxtension/ngxtension-platform/blame/991ff1e84ce06852dfd8b9ca75c135d5f883c475/libs/ngxtension/control-value-accessor/src/control-value-accessor.ts#L274 

### Example

Given you have an effect in a constructor
```
private myInput = input<string>("test");

constructor(ngxControlValueAccessor: NgxControlValueAccessor){
       effect(() => {
              const myInput = this.myInput();
              if(ngxControlValueAccessor.value){
                   // Do Something
              }
       })
}
```

### Potential Solution

In the above example it is not clear that the effect will have a dependency on myInput() as well as ngxControlValueAccessor.value$(); 
Because .value calls value$() internally. Hence either should the functional getter be untracked or not exist at all. However this is just my humble opinion and I would gladly get your inputs on this issue.

RobbyRabbitman:
fixes #547

### DIFF
--- a/libs/ngxtension/control-value-accessor/src/control-value-accessor.spec.ts
+++ b/libs/ngxtension/control-value-accessor/src/control-value-accessor.spec.ts
@@ -1,5 +1,5 @@
 import { CommonModule } from '@angular/common';
-import { Component, Provider, inject } from '@angular/core';
+import { Component, Provider, effect, inject } from '@angular/core';
 import { TestBed, fakeAsync, flush } from '@angular/core/testing';
 import {
 	FormControl,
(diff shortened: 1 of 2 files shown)


eneajaho (approved):
LGTM

PR #556, opened 2025-02-04
refactor(ngx-control-value-accessor): use normal effect instead of rxEffect

RobbyRabbitman:
rxEffect was just used in order to 'escape' setting signals in an effect back then when it was not allowed... this can be done now anyways and could have been done back then aswell, but I did not know at that time about the usage of a normal effect and untracked.

wait for #555

### DIFF
--- a/libs/ngxtension/control-value-accessor/src/control-value-accessor.ts
+++ b/libs/ngxtension/control-value-accessor/src/control-value-accessor.ts
@@ -1,6 +1,5 @@
 import {
 	Directive,
-	Injector,
 	Input,
 	Output,
 	booleanAttribute,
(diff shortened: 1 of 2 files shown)


eneajaho:
@RobbyRabbitman can you rebase this please? And make it ready for review if there's nothing else you want to add. Thanks.

libs/ngxtension/control-value-accessor/src/control-value-accessor.ts
```diff
@@ -240,30 +236,38 @@ export class NgxControlValueAccessor<T = any>
...
+	protected readonly notifyNgControlOnValueChanges = effect(() => {
+		const value = this.value$();
+
+		untracked(() => {
+			/**
+			 * no-op if THIS value is already _equal_ to the value of the control => we dont need to notify the control.
+			 */
+			if (this.compareTo(this.ngControl?.value, value)) {
+				return;
+			}
```
nartc, line 248: nit: although this does not change anything in terms of runtime but can we move this out of the `untracked` to make `untracked` usage more intentional? (as `compareTo` is already untracking the signal?)

RobbyRabbitman: yes sure, its my personal style of writing effects: read all signals you want to track, and then wrap everything with untracked so you don't accidentally read any other signals.

PR #564, opened 2025-02-16
feat: notPattern

endlacer:
### Description

Currently, Angular's form validation system provides [Validators.pattern()](https://angular.dev/api/forms/PatternValidator) to validate that a form control matches a specific regular expression pattern. However, when we need to validate that a value does NOT match a pattern, developers have to write negated regular expressions (e.g., using negative lookahead or inverse character classes).

This approach has several drawbacks:
- Writing negated regex patterns is complex and error-prone
- Negated patterns can be significantly less performant, especially for complex patterns
- The intention of the code becomes less clear and harder to maintain
- Some patterns are extremely difficult or impossible to negate correctly using regex alone

Example of current workaround:
```typescript
Validators.pattern(/^(?!.*badword).+$/) // Prevents 'badword' from appearing anywhere
```

### Proposed solution

A new validator notPattern() that validates that a form control value does NOT match the given pattern.
```typescript
// Proposed API
Validators.notPattern(pattern: string | RegExp): ValidatorFn

// Usage example
formControl = new FormControl('', [
  Validators.notPattern(/badword/), // Fails if 'badword' is found
]);
```

Benefits:
- More intuitive API for negative pattern validation
- Better performance as it doesn't require complex negated regex
- Clearer code intention
- Easier to understand


nartc:
Hi, thank you for the PR. The implementation looks good to me. However, I think `patternNot` as the name is not intuitive to parse as it puts the negation at the end of the word. Can we change the name to `notPattern` instead?

endlacer:
> Hi, thank you for the PR. The implementation looks good to me. However, I think `patternNot` as the name is not intuitive to parse as it puts the negation at the end of the word. Can we change the name to `notPattern` instead?

ok, done :)

endlacer:
> @endlacer Thanks. Can we remove the `pattern-not` entry point?

ah missed that, ty :)

nartc (approved):

nartc:
@all-contributors please add @endlacer for code

Issue #549, opened 2025-01-28
SFC Migration does not migrate styles


IgorGRSE:
Repro steps:

1. Open terminal
2. Update global angular cli up to lastest version (I tested with 19.1.4)
3. Execute `ng new angular-sfc-test`
4. Select any style type (tested with CSS and SCSS)
5. Select default options
6. Wait until preparation is done
7. Execute `ng add ngxtension`
8. Open app.component.html and delete content (it's too big for SFC default limits)
9. Execute `ng g ngxtension:convert-to-sfc`
10. Select default options
11. See that app.component.ts still use app.component.css/scss

Expected: the behavior described in the [documentation](https://ngxtension.netlify.app/utilities/migrations/sfc-migration/#:~:text=It%20will%20move%20the%20styles%20from%20the%20styleUrls%20to%20the%20styles%20property.):
- app.component.ts has moved styles in `styles` property
- removed styleUrls
- removed app.component.css/scss file, linked in styleUrls

Btw, there is a wrong code-link on the [documentation page](https://ngxtension.netlify.app/utilities/migrations/sfc-migration/)

PForever:
In fact there is a documentation issue. There is the unmentioned parameter `--move-styles` 

```typescript
if (moveStyles) {
  let styleUrl =
    decoratorPropertyName === 'styleUrl'
      ? property
          .getInitializer()
          .getText()
          .slice(
            1,
            property.getInitializer().getText().length - 1,
          )
      : '';
```

```typescript
describe('convertToSFCGenerator', () => {
  let tree: Tree;
  const options: ConvertToSFCGeneratorSchema = {
    path: 'libs/my-file.ts',
    moveStyles: true,
    maxInlineTemplateLines: 10,
    maxInlineStyleLines: 10,
  };
```

endlacer:
can be closed as completed by https://github.com/ngxtension/ngxtension-platform/pull/582

PR #582, opened 2025-03-14
Updated sfc-migration documentation with adding missed parameters

PForever:
Updated the sfc-migration documentation according to that issue #549:
Now there is the flag `--move-styles` mentioned


nartc (approved):

nartc:
Thank you

PR #586, opened 2025-03-17
Feat/add control errors in forms events

RaimonxDev:
### Description

Hello,

FormEvents is a great utility. However, I also wanted a way to determine which controls have errors and the specific error types. To achieve this, I have added controlErrors, which returns a Record<string, string>.

I have updated the tests to ensure the expected values are correctly validated.

PS: This is my first contribution to open source! 🚀

Let me know if any changes are needed.

### DIFF
--- a/apps/test-app/src/app/form-events/form-events.component.ts
+++ b/apps/test-app/src/app/form-events/form-events.component.ts
@@ -1,5 +1,5 @@
 import { AsyncPipe, JsonPipe } from '@angular/common';
-import { Component, inject } from '@angular/core';
+import { Component, inject, Injector, OnInit } from '@angular/core';
 import {
 	NonNullableFormBuilder,
 	ReactiveFormsModule,
@@ -12,6 +12,13 @@ import { allEventsObservable, allEventsSignal } from 'ngxtension/form-events';
 	standalone: true,
 	imports: [ReactiveFormsModule, JsonPipe, AsyncPipe],
 	template: `
+		<label for="control">Control</label>
+		<input placeholder="Control" [formControl]="control" />
+		<pre>{{ controlData$ | async | json }}</pre>
+		<pre>{{ controlData() | json }}</pre>
+
+		<hr />
+
 		<form [formGroup]="form">
 			<label for="firstName">First Name</label>
 			<input formControlName="firstName" name="firstName" />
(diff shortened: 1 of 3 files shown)


libs/ngxtension/form-events/src/form-events.ts
```diff
@@ -109,8 +130,20 @@ export function allEventsObservable<T>(
...
+				map(() => {
+					if (form instanceof FormGroup) {
+						return errorsEvents$(form);
+					}
+					if (form instanceof AbstractControl) {
+						return errorsEvents$(form);
+					}
+					return null;
+				}),
+			),
```
nartc, line 144: question: why do we `startWith(form)`?

libs/ngxtension/form-events/src/form-events.ts
```diff
@@ -55,6 +56,25 @@ function pristineEvents$<T>(form: AbstractControl<T>) {
 	);
 }
 
+function errorsEvents$(form: FormGroup | AbstractControl) {
```
nartc, line 59: nit: `errorsEvents$` should be Observable like other `Events$` for consistency

nartc (commented):
Thanks for the contribution. I left some comments

libs/ngxtension/form-events/src/form-events.ts
RaimonxDev: Thanks for the feedback! I have moved the logic to errorsEvent, and it now returns an observable

apps/test-app/src/app/form-events/form-events.component.ts
```diff
@@ -30,20 +55,30 @@ import { allEventsObservable, allEventsSignal } from 'ngxtension/form-events';
...
+	control = this.fb.control('', Validators.required);
+
 	formTyped$ = allEventsObservable<ReturnType<typeof this.form.getRawValue>>(
 		this.form,
 	);
 	$formTyped = allEventsSignal<ReturnType<typeof this.form.getRawValue>>(
 		this.form,
 	);
+	ngOnInit() {
+		this.controlData$ = allEventsObservable<string>(this.control);
```
michael-small, line 81: nit: IMO these can be lifted out of the `ngOnInit` into class fields w/o an injector. No need for the lifecycle hook and `!`/`any`.

michael-small (approved):
LGTM but with one nitpick about the example app. Great addition to the util.

apps/test-app/src/app/form-events/form-events.component.ts
RaimonxDev: I completely agree. I just created that example to verify that it would work with the injector. I can fix it.

PR #587, opened 2025-03-19
feat(merge-http-context): add mergeHttpContext utility function

kkachniarz220:
The mergeHttpContext function combines multiple HttpContext instances into a single context, enabling easier composition of HTTP context tokens. This utility is essential when working with multiple interceptors or when context values need to be combined from different sources.

Example usage:

```ts
import { mergeHttpContext } from 'ngxtension/merge-http-context';

@Injectable()
export class UsersService {
  getUsers() {
    return this.http.get('api/users', {
      context: mergeHttpContext(withCache(), withLogResponse()),
    });
  }
}
```

```ts
const LOG_RESPONSE = new HttpContextToken<boolean>(() => false);
export function withLogResponse(): HttpContext {
  return new HttpContext().set(LOG_RESPONSE, true);
}
```

```ts name=withCache.ts
const CACHE = new HttpContextToken<boolean>(() => false);
export function withCache(): HttpContext {
  return new HttpContext().set(CACHE, true);
}
```


libs/ngxtension/merge-http-context/src/merge-http-context.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ */
+
+export function mergeHttpContext(...contexts: HttpContext[]): HttpContext {
+	return contexts.reduce((prev, curr) => {
+		Array.from(curr.keys()).forEach((contextToken) =>
+			prev.set(contextToken, curr.get(contextToken)),
+		);
+		return prev;
+	}, new HttpContext());
```
nartc, line 17: nit/question: I'm not too familiar with `HttpContext` so I might miss something but why can't we use the first context here instead of a new `HttpContext` instance?

kkachniarz220: I opted for creating a new HttpContext instance here primarily for two reasons:

1. Enhanced Code Readability: Using a separate instance makes the flow of data more explicit and easier to follow, as each context has a clear, single responsibility.

2. Immutability Best Practices: This approach prevents any unintended modifications to the original context, which is particularly important when dealing with HTTP contexts that might be used elsewhere in the request pipeline.

While reusing the first context would technically work, creating a new instance provides better isolation and makes the code more maintainable in the long run. It's a slight trade-off between memory usage and code clarity, but in this case, I believe the benefits of clearer code organization outweigh the minimal memory overhead.

nartc (approved):

Issue #569, opened 2025-02-28
injectLocalStorage - how to deal with key based on input


jonath92:
Thank you for the great library. I have a question concerning `injectLocalStorage`. We found us a couple of times in a situation similar as that: 
```ts
class MyComponent {
  readonly projectId = input.required<string>();

  readonly scrollPosition = computed(() => 
    injectLocalStorage(`scrollPosition-${this.projectId()}`)
  );
}
``` 

This unfortunately doesn't work because we than get the error: 
```
Error: NG0950: Input is required but no value is available yet
``` 

I am not sure how to deal with that? Any help is much appreciated.

eneajaho:
Hi @jonath92 
This would be great to support, but it is hard to think of a way to support this. If you have any ideas please let me know.

pawel-twardziak:
@jonath92 I am pretty sure this error has nothing to do with `injectLocalStorage`.
The error means that no value for the input (projectId) is provided while the value is declared as required.

When the component is used, it should be `<component [projectId]="value" />` but it seems like there is `<component />` somewhere in your code

PR #593, opened 2025-04-18
feat(inject-local-storage): computed key support (#569)

MillerSvt:
Closes: #569

### DIFF
--- a/libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
+++ b/libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
@@ -33,10 +33,9 @@ export function provideLocalStorageImpl(impl: typeof globalThis.localStorage) {
  */
 export type LocalStorageOptionsNoDefault = {
 	/**
-	 *
 	 * Determines if local storage syncs with the signal.
 	 * When true, updates in one tab reflect in others, ideal for shared-state apps.
-	 * Defaults to true.
+	 * @default true
 	 */
 	storageSync?: boolean;
 	/**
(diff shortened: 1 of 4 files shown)


libs/ngxtension/inject-local-storage/src/inject-local-storage.ts
```diff
@@ -65,6 +68,36 @@ export type LocalStorageOptions<T> =
...
+interface ComputedState<T> {
+	kind: Kind.COMPUTED;
+	key: string;
+	value: T;
+}
+
+type State<T> = InitialState | ComputedState<T>;
+
+type LocalStorageSignal<T> = Signal<T> &
+	Pick<WritableSignal<T>, 'set' | 'update'>;
```
MillerSvt, line 99: TODO: Restore toReadonly method

eneajaho:
Hi @MillerSvt 
Thank you very much for this! 
I have a question. Do we want to remove the current value for the previous key? Or we leave it as is?  Maybe a way to support this would be to add some hooks in the options? 

Example:
- onBeforeKeyChange ?
- onKeyChange ?

(these are just ideas, we don't have to implement them if you think they are not necessary)

What do you think?

MillerSvt:
@eneajaho Hi!
I can't yet imagine a case where custom logic is needed before and after changing the key. However, I think it will be useful to add the `clearOnKeyChange' option, and let developers decide for themselves what they need. I think this option should only be available for overloads with computed keys, and the default value should be 'true'.

eneajaho:
> @eneajaho Hi! I can't yet imagine a case where custom logic is needed before and after changing the key. However, I think it will be useful to add the `clearOnKeyChange` option, and let developers decide for themselves what they need. I think this option should only be available for overloads with computed keys, and the default value should be `true`.

Yeah, I agree on that. As you said, it should be available  only for overloads with computed keys.

MillerSvt:
@eneajaho done

eneajaho (approved):

eneajaho:
> @eneajaho done
> 
> Off topic: When I can expect new release with my fixes for computedPrevious?

Hi, thanks. 

Regarding the question. Because the release process changed a bit in v5.0, I cannot release properly anymore. So I'm waiting for @nartc to give me a hand here 🙌

MillerSvt:
@nartc Hi. When I can expect new release?

MillerSvt:
@nartc up

MillerSvt:
@nartc up

eneajaho:
> @nartc The documentation has been updated, but it does not reflect the current reality as no release has taken place. Could you inform us when we can anticipate the new version?

Sorry for this taking so long. Managed to fix this with Chau and now we have a release with the change. https://github.com/ngxtension/ngxtension-platform/releases/tag/5.1.0

PR #608, opened 2025-06-23
fix(inject-local-storage): fix reactive purity if signals reads inside stringify function

MillerSvt:


eneajaho:
Hi @MillerSvt 
Thanks for this.

eneajaho (approved):

PR #609, opened 2025-06-23
fix(inject-local-storage): fix default value creation

MillerSvt:


MillerSvt:
> Also can we add these new additions in the docs too?

I think there is no new additions. It fixes already exists feature: `defaultValue` as factory.

PR #619, opened 2025-08-07
fix: use removeChild() loop instead of replaceChildren()

denisyilmaz:
In local development with Angular v20.2-next I was not able to use the SvgSprite Directive as it fails on the server with `element.replaceChildren is not defined`. Checking the Node.JS documentation `replaceChildren()` is only available in environments that supports DOM manipulation, such as a browser or a server-side rendering library that provides DOM-like APIs.

This fix uses a while loop to remove all children one by one ensuring its also working in such environments.


eneajaho:
Hi @denisyilmaz 
Looks like there is a linting issue. Can you fix that please?  
Thanks for the fix.

eneajaho (approved):

PR #643, opened 2025-12-05
fix(inject-local-storage): fix bug where disabling `storageSync` woul…

nstadl:
…d not write new values to local storage

# Issue

- https://github.com/ngxtension/ngxtension-platform/issues/624


LuisWolf-Snapaddy:
@eneajaho Is there an ETA when this fix will be available via npm?

PR #661, opened 2026-02-04
feat(create-notifier): add support for dependencies

endlacer:
https://github.com/ngxtension/ngxtension-platform/issues/535

Added options.deps where signals can be specified, that trigger the notifier
```ts
// Emit immediately (default behavior)
notifier = createNotifier({
	deps: [someSignal],
	depsEmitInitially: true, // starts at 1 (default)
});

// Don't emit initially
notifier = createNotifier({
	deps: [someSignal],
	depsEmitInitially: false, // starts at 0
});
```

Made it so, that the value inside notify() also increases when the signals emit

### DIFF
--- a/docs/src/content/docs/utilities/Signals/create-notifier.md
+++ b/docs/src/content/docs/utilities/Signals/create-notifier.md
@@ -68,9 +68,7 @@ You can configure a notifier to automatically increment whenever specified signa
 ```ts
 userId = signal(1);
 
-userNotifier = createNotifier({
-	deps: [this.userId],
-});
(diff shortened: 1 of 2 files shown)


endlacer:
> Hi @endlacer I was thinking if we can do something like: createNotifier.withDeps([dep1, dep2]) instead of modifying the current function. The reason is because we go from creating a signal to creating a linkedSignal and the initial function which was cheap imo now will get heavier and have more things to take care of. What do you think?

true, but i like the option-syntax more. I changed it now, check it out :)

PR #665, opened 2026-02-20
Fix/issue 662 stale docs

endlacer:
(should) fix https://github.com/ngxtension/ngxtension-platform/issues/662


endlacer:
@eneajaho 
failed bc I "touched astro.yaml":
```
The following files were touched:
- .github/workflows/astro.yml
Touched files result in the following affected projects
The following projects were touched during recent changes
```

PR #667, opened 2026-02-26
chore: remove `standalone: true` property across all components, dire…

eneajaho:
…ctives, and pipes for consistency and cleanup


eneajaho:
This should be done in small chunks. Closing for now.

PR #684, opened 2026-04-13
fix(inject-network): set init value and support SSR

### LINKED ISSUE #681: injectNetwork online initial value
`injectNetwork` `online` property initial value is set to `true`, which is incorrect if the web app is disconnected from the internet on the initial load. This could be a problem with PWA cached apps or maybe embedded WebView apps (e.g. Capacitor) where the web app assets are already available on the client.

How to reproduce:
Start your Angular app on the local server with the internet disabled. `injectNetwork().online` will be set to `true` eventhough internet connection is not available.

Current:
https://github.com/ngxtension/ngxtension-platform/blob/9c515b9613b4d1639822b112c9948a91a769b56f/libs/ngxtension/inject-network/src/inject-network.ts#L111

Proposed solution:
```ts
const online = signal(window.navigator.onLine);
```

toniskobic:
Resolves https://github.com/ngxtension/ngxtension-platform/issues/681#issue-4233117324

### DIFF
--- a/libs/ngxtension/inject-network/src/inject-network.ts
+++ b/libs/ngxtension/inject-network/src/inject-network.ts
@@ -1,5 +1,11 @@
-import { DOCUMENT } from '@angular/common';
-import { inject, signal, type Injector, type Signal } from '@angular/core';
+import { DOCUMENT, isPlatformServer } from '@angular/common';
+import {
+	inject,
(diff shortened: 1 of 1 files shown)


eneajaho (approved):
Thank you @toniskobic

PR #686, opened 2026-04-28
feat: add explicitAfterRenderEffect utility with multi-phase support …

JonataB:
…and documentation

### DIFF
--- a/docs/src/content/docs/utilities/Effects-Side-Effects/explicit-after-render-effect.md
+++ b/docs/src/content/docs/utilities/Effects-Side-Effects/explicit-after-render-effect.md
@@ -1,0 +1,77 @@
+---
+title: explicitAfterRenderEffect
+description: ngxtension/explicit-after-render-effect
+entryPoint: ngxtension/explicit-after-render-effect
+badge: stable
+contributors: ['jonata-biondi']
+---
+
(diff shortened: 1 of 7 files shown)


nartc (approved):
I'd take gemini suggestion here with a grain of salt. it's not that big of a deal imo but if you feel strongly about it, you can make the adjustment before merging. stamped

JonataB:
> I'd take gemini suggestion here with a grain of salt. it's not that big of a deal imo but if you feel strongly about it, you can make the adjustment before merging. stamped

Yes. I learn something ❤️

eneajaho:
Hi @JonataB 
Wanted to ask if this fix here https://github.com/ngxtension/ngxtension-platform/pull/688 would solve this need for you. 
Since `on` would be a general solution for all types of effects (also for computed I think).

JonataB:
> Hi @JonataB Wanted to ask if this fix here #688 would solve this need for you. Since `on` would be a general solution for all types of effects (also for computed I think).

Sorry i read only now. 
"on" solves the single-phase case:  I can use afterRenderEffect({ read: on([..], fn) }) and it works; it doesn't cover the multi-phase case. 

We have different signs. Maybe we have to change 'on'.

Also ...
image

PR #688, opened 2026-05-04
feat(reactive-on): extend usages of the reactive `on` utility

JeanMeche:
You can now use it with `computer` & `afterRenderEffect`

### DIFF
--- a/libs/ngxtension/reactive-on/src/on.spec.ts
+++ b/libs/ngxtension/reactive-on/src/on.spec.ts
@@ -1,6 +1,7 @@
 import {
 	ApplicationRef,
 	Injector,
+	afterRenderEffect,
 	computed,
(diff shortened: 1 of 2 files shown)


nartc (dismissed):
gemini suggestion seems legit. JSDoc comment with examples is helpful.

libs/ngxtension/reactive-on/src/on.spec.ts
```diff
@@ -93,9 +94,12 @@ describe(on.name, () => {
 		const log: number[] = [];
 
 		effect(
-			on({ a, b }, ({ a: valA, b: valB }) => {
```
JeanMeche, line 96: Thinking more about it, I don't think we should implicitly invoke the signals.

PR #695, opened 2026-07-03
fix(when-document-visible): improve visibility state check in partiti…

endlacer:
when-document-visible:
partition() predicate reads the global document instead of the injected/overridable one the source stream actually provides. easy ssr fix


eneajaho (approved):
Thanks

PR #696, opened 2026-07-03
feat(connect): enhance reducer behavior for falsy values in signal up…

endlacer:
`reducer?.(prev, x) || x` is meant to say "if a reducer was provided, use its result; otherwise use the raw emitted value x." But || only checks truthiness, not whether the reducer ran. 
So if the reducer is defined and legitimately computes a falsy value (`0, '', false, null, ect`) the || treats that as "nothing," discards it, and silently substitutes the raw x from the observable instead.


eneajaho (approved):
Thanks

PR #697, opened 2026-07-03
fix(inject-route-fragment): return default values even when falsy

endlacer:
defaultValue used a truthy check, so falsy defaults (0, '', false) were silently discarded and `null` was returned instead.


eneajaho (approved):
Thanks

